PR #3, opened 2022-08-10
Removed group block that was creating unnecessary padding

### LINKED ISSUE #2: Root padding is applied twice in the Query Block
It looks like the root padding is applied twice in a query block:

image

image

pbking:
Fixes: https://github.com/WordPress/twentytwentythree/issues/2

This change removes the `wp:group` block from the `wp:query` block eliminating the duplicate gap.

Before:
image

After:
image


mikachan (approved):
Looks good to me, thank you!

PR #11, opened 2022-08-10
Fix order of style variations

### LINKED ISSUE #8: Reorder style variation examples
It would be nice to browse the style variations in the correct order. That would be left to right: Default > Example 2 > Example 3 > Example 4.

Also, there's a typo in "Example 2".

https://user-images.githubusercontent.com/65220155/183970398-12ab9b94-b4b5-4910-ae61-b9a2d449cbd4.mp4

mikachan:
This PR renames the variation files so that they use numbers rather than words (e.g. '2' instead of 'two'). This means that they will be listed in the correct order when viewed in Global Styles (as they're listed alphabetically).

This also fixes a typo in Example 2.

Fixes https://github.com/WordPress/twentytwentythree/issues/8.


pbking (approved):
Looks Good to me

beafialho (approved):

Issue #12, opened 2022-08-10
New font: IBM Plex Mono


critterverse:
Can we add the typeface IBM Plex Mono to the font assets? Thank you!

colorful-tones:
@pbking @critterverse  according to my quick research with [Font Scanner plugin](https://www.figma.com/community/plugin/1002875535237160651/FontScanner) in Figma against the TT3 file. IBM Plex Mono: Regular, Light and Bold were used. Hence, I included: Bold, Italic, Regular and Light in #17 . Granted, there is a lot going on in #17 😄 

I _do_ think a larger discussion needs to be had about including fonts and this will impact design. Choosing fonts is nuanced and performance should be considered, which is not always easy for designer to consider. Variable fonts are great, but it is not a clear win in a lot of cases. Here is a simple overview of the issue: [Font Techniques for Modern WordPress Themes](https://wpengine.com/builders/font-techniques-wordpress-themes/)

So, considerations of:

1. Pick only the bare minimum of weights and styles needed
2. Based on the weights and styles: determine if static or variable font is beneficial
3. Only bundle based on first two ☝️ 

I would also encourage us to refrain from loading too many of italic styles, because the browser does a pretty good job of inferring italic without an explicit font file reference. I tend to only include the regular (400) italic as a minimum. 🤔 

<img width="595" alt="Screen Shot 2022-08-11 at 7 04 44 AM" src="https://user-images.githubusercontent.com/405912/184120126-dde73b3b-a96b-404b-9381-9eea14e59a4a.png">

pbking:
I agree wholeheartedly with the above. 

There's also the potential of dynamically loaded fonts (Google Fonts and the like).  Should that be an option the entire font opinion can be an option in each variation.  I'm very keen to see this solution available.

Regardless, the theme's font selection has to be well curated.

Should we have a target total filesize for each?  "As small as possible" can still vary quite a bit across typefaces.

carolinan:
if only the font weights used by the design are included, that can impact the sites when users changes the text appearance.

pbking:
> if only the font weights used by the design are included, that can impact the sites when users changes the text appearance.

That's a good point @carolinan.  It's important to keep the payload small, but also to empower not only the variation designers but also the user who will have no control of what weights are available.

[...]

Issue #10, opened 2022-08-10
Example Variations Using Different Fonts


bgardner:
Given the likelihood that variations from community authors will use different Google fonts, at least one of the example variations should demonstrate how it is loaded in its theme.json file such as:

```{
	"title": "Example X",
	"settings": {
		"color": {
			"palette": [
				{
					"color": "#ffd875",
					"name": "Base",
					"slug": "base"
				},
				{
					"color": "#000000",
					"name": "Contrast",
					"slug": "contrast"
				},
				{
					"color": "#ffffff",
					"name": "Primary",
					"slug": "primary"
				},
				{
					"color": "#000000",
					"name": "Secondary",
					"slug": "secondary"
				},
				{
					"color": "#F6F6F6",
					"name": "Tertiary",
					"slug": "tertiary"
				}
			]
		},
		"typography": {
			"fontFamilies": [
				{
					"fontFace": [
						{
							"fontFamily": "Open Sans",
							"fontStretch": "normal",
							"fontStyle": "normal",
							"fontWeight": "200 900",
							"src": [
								"file:./assets/fonts/open-sans/OpenSans-Regular.ttf"
							]
						},
						{
							"fontFamily": "Open Sans",
							"fontStretch": "normal",
							"fontStyle": "italic",
							"fontWeight": "200 900",
							"src": [
								"file:./assets/fonts/open-sans/OpenSans-Italic.ttf"
							]
						}
					],
					"fontFamily": "\"Open Sans\", sans-serif",
					"name": "Open Sans",
					"slug": "open-sans"
				}
			]
		}
	},
	"styles": {
		"blocks": {
			"core/separator": {
				"border": {
					"width": "2px"
				}
			}
		},
		"elements": {
			"button": {
				"border": {
					"radius": "5px"
				}
			}
		},
		"typography": {
			"fontFamily": "var(--wp--preset--font-family--open-sans)"
		}
	},
	"version": 2,
	"$schema": "https://schemas.wp.org/trunk/theme.json"
}```

pbking:
This would be a good way to solve #12.

mikachan:
We're not currently planning on using any fonts other than those bundled with the theme, so I'm going to close this for now.

PR #17, opened 2022-08-10
Add IBM Plex Mono Typeface and added it to a variation as an example

pbking:
Not sure if we want this ONLY in a variation (or specifically that variation), but doing so addresses #10.

Adding the font addresses #12 (but only for that variation).


mikachan (commented):
Thanks for adding this font!

I think style variation 5 was intended to use 'Source Serif Pro', according to [the designs here](https://www.figma.com/file/j469TEkvf2YJ3EVQm9X4Nt/TT3?node-id=253%3A7718). However, variation 2 uses a mono font globally, and variation 4 uses a mono font for paragraphs, so 'IBM Plex Mono' should probably be used in these cases.

> Not sure if we want this ONLY in a variation (or specifically that variation)

I believe the intention was for TT3 to only be bundled with 3-4 font families (see [the project kick-off post](https://make.wordpress.org/design/2022/08/10/twenty-twenty-three-default-theme-project-kickoff/)), so I think it would be best if they are all defined in the main theme.json file, and then the variations can use the font `slug` to reference the desired font.

pbking:
Closing in lieu of #65

PR #18, opened 2022-08-10
Replace PHP minimum version

audrasjb:
Because PHP 5.7 never existed :)
https://www.php.net/releases/index.php#5.6.40

### DIFF
--- a/style.css
+++ b/style.css
@@ -6,7 +6,7 @@ Author URI: https://wordpress.org
 Description:
 Requires at least: 5.9
 Tested up to: 6.1
-Requires PHP: 5.7
+Requires PHP: 7.4
 Version: 0.0.1
 License: GNU General Public License v2 or later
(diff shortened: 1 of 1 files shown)


carolinan (approved):
7.4 is the recommended version for WP so makes sense.
(It was decided to not increase the recommended PHP version to 8.0 in 6.1)

thelovekesh (approved):

mikachan (approved):

style.css
```diff
@@ -6,7 +6,7 @@ Author URI: https://wordpress.org
 Description:
 Requires at least: 5.9
 Tested up to: 6.1
-Requires PHP: 5.7
+Requires PHP: 7.4
```
Soean, line 9: WordPress 6.1 has the minimium PHP version 5.6, so you should be able to install the theme with this version.

```suggestion
Requires PHP: 5.6
```

PR #19, opened 2022-08-10
Replace license link in style.css

audrasjb:
Because it doesn't make much sense to point to an Automattic URL on GitHub :)


mikachan (approved):
Ah good spot, thank you!

Issue #22, opened 2022-08-11
Use generic font family slugs


richtabor:
We should consider using more generic font-family slugs (i.e. `body`, `heading`, etc) instead of slugs that are specific to the font family that is registered within `settings.typography.fontFamilies`. 

The fonts loaded within `settings.typography.fontFamilies` are only available on the active style, resulting in CSS variables 

### Why: 

If a the parent theme.json file registers a font, like DM Sans, and uses a slug of `dm-sans` — whenever we switch styles we're left with an undefined CSS variable output wherever we used `"fontFamily": "var(--wp--preset--font-family--system-font)"`. 

Also resulting in none of the parent theme.json's font family values will carry over to subsequent styles. 

Below is a screenshot of the CSS generated if you have a specific font-family slug from the parent theme.json that is not available (different font) from a style variation.

### Proposal:
Instead if we used `body` and `heading` slugs, then those CSS variables will be meaningful within other style variations. A heading with `"fontFamily": "var(--wp--preset--font-family--heading)",` assigned to it, still maintains the style variant's fontFamily that has the slug of `heading`.

Having a style that actually inherits the parent style seems the best foot forward in my opinion. And even if we end up not having fontFamilies within the parent theme.json, we should still have these consistent values across the theme — from variation to variation. Example: if I choose my theme's heading font (DM Sans) then switch styles, then that style variation's heading font should take over wherever I used the previous one. 

What do you think?

### Screenshot: 
<img width="695" alt="CleanShot 2022-08-10 at 22 01 20@2x" src="https://user-images.githubusercontent.com/1813435/184052663-5a3ba8bc-b197-4e17-9b01-819b0f145cdb.png">
<img width="495" alt="CleanShot 2022-08-10 at 21 45 55@2x" src="https://user-images.githubusercontent.com/1813435/184052362-f45d8d9a-1397-460e-a62d-485afea631ea.png">

colorful-tones:
I support this endeavor! 💯 🥳

pbking:
I'm just [throwing spaghetti](https://idioms.thefreedictionary.com/throw+spaghetti+at+the+wall+and+see+what+sticks) here;

What if we both allow users to use any of the fonts included while also allowing the variation to dictate which go where by default?  

Say we include all available fonts in the base theme.  Therefore, any variation would have a specific font should the user choose one.

[...]

Issue #9, opened 2022-08-10
Style variation thumbnails: Not displaying fonts used


critterverse:
The style variation thumbnail previews don't currently reflect the fonts used in each variation. The smaller Aa Aa preview that appears on hover looks correct, but the larger text on hover and the Aa text in the main preview are not displaying properly:

![variation-thumbnails](https://user-images.githubusercontent.com/77359364/183973454-83398060-25d0-47c6-ab91-9b71f6f08354.jpg)

bgardner:
@critterverse Do you have Gutenberg active? I have experienced this same thing, but only with WP core. When I activate Gutenberg, this is resolved.

mikachan:
It looks like this has been fixed, as I can see the fonts here now (I couldn't before!)

<img width="273" alt="image" src="https://user-images.githubusercontent.com/1645628/185373091-e939613e-8b18-406a-a008-a3ab1ead6a11.png">

I'm guessing there must have been a fix upstream in GB. If we notice it again we can just open this issue again.

PR #23, opened 2022-08-11
Propose new font slug tokenization and expose in Example Two

### LINKED ISSUE #12: New font: IBM Plex Mono

colorful-tones:
This mostly addresses #22 , fixes #12 and addresses #10. Oh, and mostly addresses #9 (for Example 2 at least). 😅 

When I implemented style variations in [Extendable](https://github.com/extendify/extendable) this was the best solution for incorporating fonts in variations. As @richtabor noted in #22 . We'll want to leverage font slugs, e.g. `headings` and `body` and then we can customize in each variation.

It is crucial to keep in mind that _anything_ set in the main `theme.json` will have to be unset or neutralized in a style variation. A good example is how currently [`h1`](https://github.com/WordPress/twentytwentythree/blob/trunk/theme.json#L276), [`h2`](https://github.com/WordPress/twentytwentythree/blob/trunk/theme.json#L284), [`h3`](https://github.com/WordPress/twentytwentythree/blob/trunk/theme.json#L292) all have explicit `font-weight: 300` set in the `theme.json`. Every style variation will have to set something else unless they _all_ want to have `font-weight: 300` for those headings as well.


theme.json
```diff
@@ -200,9 +170,9 @@
 			},
 			"core/post-title": {
 				"typography": {
-					"fontFamily": "var(--wp--preset--font-family--system-font)",
+					"fontFamily": "var(--wp--preset--font-family--headings)",
```
richtabor: You have the slug set to `heading` up above, I think we should use that nomenclature, instead of headings

theme.json
```diff
@@ -68,6 +68,21 @@
 				}
 			]
 		},
+		"custom": {
+			"typography": {
+				"font-weight": {
```
richtabor, line 73: I actually don't think there's a lot of value added by separating font weight within custom, as these values are "set in stone". For example, extraLight will always have a font-weight of 200. I'd rather us use the actual values throughout instead.

colorful-tones: Hmm, I think we should clarify "set in stone." Because, variable fonts allow for values like `210`, `450`, etc.

theme.json
colorful-tones: Great catch! I fixed that up.

theme.json
colorful-tones: This system sets us up so that style variations could potentially set their own `custom.typography.font-weight.regular` to `430` if they desire. 🤔

bgardner: @colorful-tones Worth noting that in cases where variable numbers are set in theme.json `210`, `450`, etc as you have indicated, those are not used when someone uses Appearance in the block editor to change weight. IOW, regardless of what is defined in custom., there is conflicting output and hardcoded rounded values are used: `200`, `300`, `400`, etc

bgardner: I know this because Outfit is generally a thicker font, so I tried setting custom weights in theme.json to `150`, `250`, `350`, `450`, etc and those were ignored when I selected weights via the Appearance in editor. Medium was hardcoded inline as `500`, even though I had it at `450` in theme.json.

colorful-tones: Yes, I can see how the current implementation in Gutenberg is limiting and only allows the UI to show and set traditional values.

colorful-tones:
I will likely close this out by the end of the day because I feel it is possibly causing more confusion or noise than moving things forward. Perhaps I tried to roll in too many fixes into a single PR, which likely causes some mental dissonance for reviewers.

pbking:
@colorful-tones I rolled only the fonts into the main theme.json [here](https://github.com/WordPress/twentytwentythree/pull/65 ).

PR #27, opened 2022-08-11
Remove site logo and wrapping row

### LINKED ISSUE #7: Base theme: remove site logo
It would be nice to remove the site logo from the default header, so we can have this instead:

Captura de ecrã 2022-08-10, às 17 49 45

carolinan:
Removes the site logo. Closes https://github.com/WordPress/twentytwentythree/issues/7.

Also removes the group block that was only used to group the logo and title together to place them both to the left of the menu, because it is no longer needed.


mikachan (approved):
LGTM, thank you!

PR #29, opened 2022-08-11
Add CS and linting with required GitHub workflows

### LINKED ISSUE #28: Add code linting with required GitHub workflows
Add code linting to make sure WordPress Coding Standards are maintained during the development.

Tasks:

- [x]  Add Stylelint for CSS linting.
- [x]  Add PHPCS for PHP CS and linting.
- [x]  Add required GitHub workflows to run linting on PRs.

Need Discussion:
- Add `pre-commit` hook
- Is config for JS linting required?

thelovekesh:
Fixes: #28


carolinan:
Is the theme even going to use any other CSS than the font smoothing?

thelovekesh:
@carolinan 👋🏼 

Since the theme contains some CSS, IMO there should be a way to confirm that things follow WPCS everywhere.

After setting up PHPCS I also found some fixable warnings. Can you please check and confirm them?

thelovekesh:
@colorful-tones Can you please assign reviewers to this PR?

colorful-tones:
@thelovekesh I've assigned the project leads for code review. I've also labelled this Issue.

.nvmrc
```diff
@@ -0,0 +1 @@
+14
```
mikachan, line 1: Should we use version 14, or the latest version (e.g. `lts/*`)? I don't think it will matter too much, but I wondered if there was a reason we shouldn't use the latest version here.

package.json
```diff
@@ -0,0 +1,30 @@
...
+		"WordPress",
+		"theme"
+	],
+	"homepage": "https://github.com/WordPress/twentytwentythree#readme",
+	"repository": {
+		"type": "git",
+		"url": "git+https://github.com/WordPress/twentytwentythree.git"
+	},
+	"bugs": {
+		"url": "https://github.com/WordPress/twentytwentytwo/issues"
```
mikachan: This URL should be for `twentytwentythree`.

mikachan (commented):
This looks great to me, thank you for working on it!

I've left a few inline comments, but nothing major.

I agree with @carolinan that we're probably not going to add any more CSS, but I understand your point that if the theme contains any CSS, we should have a way to confirm we're following the correct standards.

I saw the fixable warnings when I ran the script. I believe they're correct according to the standards.

.nvmrc
thelovekesh: I keep the NodeJS version same as the Gutenberg repo.

mikachan: Ah OK, that sounds fine to me 👍

mikachan (approved):

PR #31, opened 2022-08-11
Remove opinion from templates

scruffian:
There are quite a lot of styles in the templates which I think can be achieved in theme.json. If we remove these then it gives us a greater range of style options we can achieve using theme.json.

Where it makes sense I have added the opinion to theme.json.

### DIFF
--- a/templates/search.html
+++ b/templates/search.html
@@ -1,21 +1,21 @@
 <!-- wp:template-part {"slug":"header","tagName":"header"} /-->
 
-<!-- wp:group {"style":{"spacing":{"padding":{"bottom":"var(--wp--preset--spacing--90)"}}},"layout":{"inherit":true}} -->
-<div class="wp-block-group" style="padding-bottom:var(--wp--preset--spacing--90)"><!-- wp:search {"label":"Search","showLabel":false,"width":100,"widthUnit":"%","buttonText":"Search","buttonUseIcon":true} /--></div>
+<!-- wp:group {"layout":{"inherit":true}} -->
+<div class="wp-block-group"><!-- wp:search {"label":"Search","showLabel":false,"width":100,"widthUnit":"%","buttonText":"Search","buttonUseIcon":true} /--></div>
 <!-- /wp:group -->
 
 <!-- wp:query {"query":{"perPage":10,"pages":0,"offset":0,"postType":"post","categoryIds":[],"tagIds":[],"order":"desc","orderBy":"date","author":"","search":"","exclude":[],"sticky":"","inherit":true},"layout":{"inherit":true},"tagName":"main"} -->
 <main class="wp-block-query"><!-- wp:post-template {"align":"wide"} -->
 <!-- wp:group {"layout":{"inherit":true}} -->
-<div class="wp-block-group"><!-- wp:post-title {"isLink":true,"align":"wide","fontSize":"x-large"} /-->
+<div class="wp-block-group"><!-- wp:post-title {"isLink":true,"align":"wide"} /-->
 
-<!-- wp:post-featured-image {"isLink":true,"align":"wide","style":{"spacing":{"margin":{"top":"calc(1.75 * var(--wp--style--block-gap))"}}}} /-->
+<!-- wp:post-featured-image {"isLink":true,"align":"wide"} /-->
 
 <!-- wp:columns {"align":"wide"} -->
 <div class="wp-block-columns alignwide"><!-- wp:column {"width":"650px"} -->
 <div class="wp-block-column" style="flex-basis:650px"><!-- wp:post-excerpt /-->
 
-<!-- wp:post-date {"isLink":true,"format":"F j, Y","style":{"typography":{"fontStyle":"italic","fontWeight":"400"}},"fontSize":"small"} /--></div>
+<!-- wp:post-date {"isLink":true,"format":"F j, Y"} /--></div>
 <!-- /wp:column -->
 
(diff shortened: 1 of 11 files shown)


templates/single.html
```diff
@@ -23,13 +23,13 @@
...
-<div class="wp-block-group"><!-- wp:post-date {"format":"F j, Y","style":{"typography":{"fontStyle":"italic","fontWeight":"400"}},"fontSize":"small"} /-->
+<div class="wp-block-group"><!-- wp:post-date {"format":"F j, Y"} /-->
 
-<!-- wp:post-author {"showAvatar":false,"fontSize":"small"} /-->
+<!-- wp:post-author {"showAvatar":false} /-->
 
-<!-- wp:post-terms {"term":"category","fontSize":"small"} /-->
+<!-- wp:post-terms {"term":"category"} /-->
 
-<!-- wp:post-terms {"term":"post_tag","fontSize":"small"} /--></div>
```
MaggieCabrera, line 32: this looks like the perfect candidate to add to theme.json

templates/single.html
```diff
@@ -2,9 +2,9 @@
 
 <!-- wp:group {"tagName":"main"} -->
 <main class="wp-block-group"><!-- wp:group {"layout":{"inherit":true}} -->
-<div class="wp-block-group"><!-- wp:post-title {"level":1,"align":"wide","style":{"spacing":{"margin":{"bottom":"var(--wp--preset--spacing--70)"}}}} /-->
+<div class="wp-block-group"><!-- wp:post-title {"level":1,"align":"wide"} /-->
 
-<!-- wp:post-featured-image {"align":"wide","style":{"spacing":{"margin":{"bottom":"var(--wp--preset--spacing--70)"}}}} /-->
```
MaggieCabrera, line 7: Why remove this one but it's fine on others like `footer-default.html`?

scruffian: I feel like this should be handled by block gap with the flow of the document, whereas the header and footer are outside that flow.

templates/single-no-separators.html
```diff
@@ -15,13 +15,13 @@
 
 <!-- wp:group {"layout":{"inherit":true}} -->
 <div class="wp-block-group"><!-- wp:group {"layout":{"type":"flex"}} -->
-<div class="wp-block-group"><!-- wp:post-date {"format":"F j, Y","style":{"typography":{"fontStyle":"italic","fontWeight":"400"}},"fontSize":"small"} /-->
+<div class="wp-block-group"><!-- wp:post-date {"format":"F j, Y"} /-->
```
MaggieCabrera, line 18: should be in theme.json too imo

jffng:
> Where it makes sense I have added the opinion to theme.json.

I think this is missing from the PR?

scruffian:
Rebased from trunk. I think this is ready for another review.

jffng (commented):
The only change I noticed was is in the post titles in the single + page templates, they are smaller now: 

Before | After
--- | ---
Screen Shot 2022-08-11 at 3 37 50 PM | Screen Shot 2022-08-11 at 3 38 04 PM

If we're okay with that, this PR LGTM since it will allow for more design flexibility in the variations.

scruffian:
I made the font size bigger on the Post Title. I also added a new xx-large font size, so that people are following the right naming conventions for fonts.

I also enabled fluid typography in the base theme as I feel like that's a good default to have turned on...

templates/search.html
```diff
@@ -1,21 +1,21 @@
...
+<div class="wp-block-group"><!-- wp:search {"label":"Search","showLabel":false,"width":100,"widthUnit":"%","buttonText":"Search","buttonUseIcon":true} /--></div>
 <!-- /wp:group -->
 
 <!-- wp:query {"query":{"perPage":10,"pages":0,"offset":0,"postType":"post","categoryIds":[],"tagIds":[],"order":"desc","orderBy":"date","author":"","search":"","exclude":[],"sticky":"","inherit":true},"layout":{"inherit":true},"tagName":"main"} -->
 <main class="wp-block-query"><!-- wp:post-template {"align":"wide"} -->
 <!-- wp:group {"layout":{"inherit":true}} -->
-<div class="wp-block-group"><!-- wp:post-title {"isLink":true,"align":"wide","fontSize":"x-large"} /-->
+<div class="wp-block-group"><!-- wp:post-title {"isLink":true,"align":"wide"} /-->
 
-<!-- wp:post-featured-image {"isLink":true,"align":"wide","style":{"spacing":{"margin":{"top":"calc(1.75 * var(--wp--style--block-gap))"}}}} /-->
```
pbking, line 12: Nearly all of the featured image blocks are losing a top margin.

Would it be proper to define that in theme.json?

I don't think that needs to be decided in this change... Removing that from the template definitely feels important.

pbking: nvm, that feels like flavor for a variation.  I would argue the stock implementation shouldn't.

pbking:
There are a few 32px spacers scattered about.  Could those be instead tied to a preset spacing value?  (i.e. `--wp--preset--spacing--90` or something).

That's the only thing I see that we might yet want to move from template to configuration and shouldn't delay bringing this in.

These are good changes and it's good to see much less opinion in the templates.

scruffian:
I don't think its possible to use variables for spacers.

mikachan (approved):
Thanks for all the work on this! I agree with the above comments, it's great to see all these settings moved to theme.json.

We could revisit the spacers in another PR if needed.

PR #32, opened 2022-08-11
Remove page numbers

scruffian:
I don't think page numbers add much value, and they look really bad on the first and last pages, so I think we should just remove them!


mikachan (commented):
This makes sense to me and matches what's in the [Figma mockups](https://www.figma.com/file/j469TEkvf2YJ3EVQm9X4Nt/TT3?node-id=147%3A1719).

Is this OK with you, @beafialho?

beafialho:
Yes, totally agree!

mikachan (approved):

PR #33, opened 2022-08-11
Change PHP minimum version to 5.6

Soean:
WordPress 6.1 has the minimium PHP version 5.6, so you should be able to install the theme with this version.

It was changed in #18.


audrasjb:
Yeah, 5.6 sounds good to me.

mikachan (approved):
Thanks!

Issue #1, opened 2022-08-10
style.css and functions.php are only required to handle font-smoothing


mikachan:
Currently, the style.css and functions.php files are only required to add the following CSS:

```
body {
	-moz-osx-font-smoothing: grayscale;
	-webkit-font-smoothing: antialiased;
}
```

I don't believe we'll need to add any other CSS, so it's a shame to include these files just to handle font-smoothing. Unfortunately, the current system font we're using looks significantly worse without the above CSS. 

A solution could be to choose a different default font, but that may require bundling more font files. 

Another idea could be to add a theme.json setting that opts-in to font-smoothing, however this has been discussed previously: https://github.com/WordPress/gutenberg/issues/35934

Would love to hear any more ideas.

bgardner:
I use font-smoothing on every theme/site I design—and that is many of them. I was bummed to see the initial issue closed because I think the move to css-less building is a good one. And I feel you @mikachan, loading a style sheet just for font-smoothing is hard to digest, but given my love for font-smoothing, is something I could get behind.

pbking:
Better to load the CSS than choose an alternative font. Otherwise other fonts added by variations would have the same limitation.  

To that end [one of the comments against](https://github.com/WordPress/gutenberg/pull/36092#issuecomment-957401611):
> We use theme.json for two things, essentially:
> * to be able to easily access CSS values for displaying them in the matching UI controls
> * to coordinate values that come from different sources (core, theme, user) so we only enqueue one of them
>
> It looks like this property doesn't check any of the boxes. 

doesn't seem to fit the bill any more.  A variation or child theme may want a different font-smoothing configuration based on font sizes chosen or typefaces provided.  This seems to make it a good reason to provide the option in theme.json.

For example, if a variation leverages "8-bit style fonts" or something that doesn't render well with smoothing there's no option to disable it. 

I think it makes sense to revisit [36092](https://github.com/WordPress/gutenberg/pull/36092) again.

richtabor:
Agreed. Something like `settings.typography.fontSmoothing` to `true`.

scruffian:
Here are two good articles about the problems with font smoothing:
https://usabilitypost.com/2012/11/05/stop-fixing-font-smoothing/
https://ui.dev/rwd/articles/font-smoothing

IMO we should remove this CSS, it's unhelpful and all other themes will follow it.

PR #37, opened 2022-08-11
Try to improve default system font appearance on macOS

mikachan:
This PR attempts to address the concerns raised in https://github.com/WordPress/twentytwentythree/issues/1, without requiring a new theme.json setting for font-smoothing. I mainly want to demonstrate what is currently possible without font-smoothing.

In order to remove the need for any CSS (and therefore also the need for a functions.php to enqueue the stylesheet), I've applied a default font-weight of `300` to everything, which I think is as close as we can get using the current font stack compared to using font-smoothing.

With font-smoothing (current implementation):
image

Without font-smoothing and with font-weight: 300 (this PR):
image

Without font-smoothing and a default font-weight (this is the original state that font-smoothing is addressing):
image

The work in this PR is just one attempt, here are some alternatives:

### Remove `-apple-system` and `BlinkMacSystemFont` from the font stack

This means that on macOS, the font would default to 'Helvetica Neue':

image

### Use `Inter` instead

@beafialho suggested `Inter` as an alternative font:

image

This looks great but would mean the theme would need to include additional font files. If we include all weights and styles, this is an additional file size of 805KB (based on TT2's Inter font file). Or, if we choose a subset of styles and weights, we could probably significantly reduce this file size (depending on what we choose). The draw-back of this solution is that it would mean there is no font option included in the theme that doesn't require font files to be loaded, which is a massive benefit to the current default font option.


pbking:
I think there should be a system-only Font Family available.
Including a "(System)" Font family doesn't cost any space.
Maybe it's not the default and looks strikingly similar to 'Inter'.  But It doesn't require font files and that might be an important option.

beafialho:
From all these alternatives, the ones that look better are `Inter` and the system font with `font-smoothing`.

_**"Without font-smoothing and a default font-weight"**_ doesn't look great, and we should definitely rule out Helvetica.

It's definitely not an ideal solution, but if we need to compromise we could probably go with **_"Without font-smoothing and with font-weight: 300"_** but would need to adjust letter spacing as it looks tight.

mikachan:
Thanks @beafialho! That helps a lot, as I think that narrows our choices down to:

1. System font with font-smoothing, as we have currently
2. System font without font-smoothing, with a font-weight of 300 and adjustments to letter spacing

I'd love to be able to remove all CSS, but I understand it's not ideal. It would be great to hear some other opinions. I'm happy for this PR to stay open if we want to discuss this further, as I don't think the code changes will cause any complicated conflicts in the future (unless we add any more CSS!)

beafialho:
> I'd love to be able to remove all CSS, but I understand it's not ideal. It would be great to hear some other opinions.

I'd love to hear other opinions as well!

colorful-tones:
I would vote that we just leave things as is for `font-smoothing` in `style.css`. I'm not a big fan of `font-weight: 300` and less a fan of `letter-spacing`.

Mostly, my reasoning is that many developers look to core theme for best practices. Therefore, I would hate for `font-weight: 300` to become some misunderstood best practice for all themes. 🤔 

Of course, if we want to pursue [this approach](https://github.com/WordPress/twentytwentythree/issues/22#issuecomment-1213000974) for the overall project then I would say breaking the `font-smoothing` out into a custom property in [the `theme.json`](https://github.com/extendify/extendable/blob/main/theme.json#L159-L162) would be ideal, and reference [in the `style.css`](https://github.com/extendify/extendable/blob/main/style.css#L29-L32). But, for the current project's infrastructure we can just avoid integrating custom property in `theme.json`.

colorful-tones:
@mikachan I think we're good to close this based on #155 right?

mikachan:
Yes, I think so, thanks for the nudge @colorful-tones!

PR #39, opened 2022-08-12
Clean base theme theme.json file

mikachan:
This PR cleans up the base theme.json file.

- Removed a reference to a template that doesn't exist (page large header)
- Removed multiple references to the global font family, as they're unnecessary 
- Moved repeated settings from individual headings to the heading element
- Set the navigation font size to `small`

### DIFF
--- a/theme.json
+++ b/theme.json
@@ -301,29 +301,23 @@
 			"h2": {
 				"typography": {
 					"fontSize": "clamp(2.625rem, calc(2.625rem + ((1vw - 0.48rem) * 8.4135)), 3.25rem)",
-					"fontWeight": "300",
 					"lineHeight": "1.2"
 				}
 			},
 			"h3": {
 				"typography": {
-					"fontSize": "2.25rem",
-					"fontWeight": "300",
-					"lineHeight": "1.4"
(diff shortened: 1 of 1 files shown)


theme.json
```diff
@@ -246,7 +241,6 @@
 			},
 			"core/query-title": {
 				"typography": {
-					"fontFamily": "var(--wp--preset--font-family--system-font)",
 					"fontSize": "clamp(3.25rem, calc(3.25rem + ((1vw - 0.48rem) * 8.4135)), 3.625rem)",
```
scruffian: I think we should just remove this

theme.json
```diff
@@ -246,7 +241,6 @@
...
 					"fontWeight": "300",
```
scruffian: And this

theme.json
```diff
@@ -246,7 +241,6 @@
...
 					"lineHeight": "1.2"
```
scruffian: And this

mikachan: Thanks! Those settings weren't making any difference. I've removed them.

scruffian (approved):
LGTM

PR #41, opened 2022-08-12
Try a simpler home template

scruffian:
The home template contains a 650px wide column, which is a setting from theme.json. This PR tries to remove that static width and instead replace it with inheriting the default layout. The downside is that now the content is no longer offset:

Before:
Screenshot 2022-08-12 at 14 40 09
Screenshot 2022-08-12 at 14 41 34

After:
Screenshot 2022-08-12 at 14 39 34

Screenshot 2022-08-12 at 14 41 47

I'm not sure if this is an improvement or not.


carolinan:
I have a _slight_ preferation for the first. There is something about the title not lining up with the image.
Compared to the 3 columns design, I prefer the 3 columns.

mikachan:
I agree with @carolinan, I also prefer the first as the title and image alignment don't look quite right.

The 3 column design is being implemented here: https://github.com/WordPress/twentytwentythree/pull/46

mikachan:
I'm going to close this as we've merged the 3 column layout (https://github.com/WordPress/twentytwentythree/pull/46) for the home template.

But I still think it's worth exploring the removal of the static 650px column width. We currently still use this in the index template.

PR #42, opened 2022-08-12
Spacers: Remove static pixel value

scruffian:
This removes static pixel values from spacers so they can be controlled by theme.json


mikachan (approved):
Will be great to control these from theme.json, thank you!

Issue #43, opened 2022-08-12
Home template: match the Figma


jffng:
Currently the home template is a query block list-style view. It should match the Figma mockup:

![home](https://user-images.githubusercontent.com/5375500/184420034-60f958c9-d44e-4e10-ba23-4220871fc3e3.png)

carolinan:
<del>1. Is there a design also for the paginated result? -I would like to know if the image and heading is meant to show on page two, or if they outside the query.</del>
Without adding CSS:
-There is no way to hide content on the paginated results.
-Looking at the archive page, the pagination is below the posts, -it is not possible to hide this pagination conditionally.

2. There is no **valid** way to place the query pagination block next to the "Latest posts" text, because its only valid parent is the query loop. There is a work around, but it also means that users who does not know HTML (or block markup) can not replicate this design themselves.

carolinan:
If both the blog heading "mindblown" and the "latests posts", are headings -and they probably should be... then we end up with the post titles being H4. That might be correct, but it sort of goes against the gut feeling, because I think these titles are more important.

beafialho:
We can perhaps simplify and remove "Latest Posts":

<img width="542" alt="Captura de ecrã 2022-08-22, às 10 01 49" src="https://user-images.githubusercontent.com/65220155/185882975-33a6acee-445b-4489-b0c5-029eeb01fbab.png">

carolinan:
I tried using [is_paged()](https://developer.wordpress.org/reference/functions/is_paged/) in a pattern to output different query paginations on the home page and page two (/page/2/), but it did not work, it always returns false.
That means that page 2+ would have the same text "Read All Posts" and that is why I recommend using the same pagination as on the archive. _Or no pagination... but I think that would potentially be unexpected and confusing to users._

beafialho:
Good point! We can have pagination:

<img width="593" alt="Captura de ecrã 2022-08-23, às 08 39 56" src="https://user-images.githubusercontent.com/65220155/186100372-607140b0-b1eb-446c-a6e9-47c9797f8006.png">

PR #46, opened 2022-08-15
Update home template

carolinan:
Update the home template to use three columns of blog posts, + an image with a heading below it.
For https://github.com/WordPress/twentytwentythree/issues/43.


templates/home.html
```diff
@@ -1,35 +1,44 @@
...
-
-	<!-- wp:query-pagination {"paginationArrow":"arrow","align":"wide","layout":{"type":"flex","justifyContent":"space-between"}} -->
-	<!-- wp:query-pagination-previous /-->
-
-	<!-- wp:query-pagination-numbers /-->
-
-	<!-- wp:query-pagination-next /-->
-	<!-- /wp:query-pagination --></main>
+<!-- wp:group {"tagName":"main","style":{"spacing":{"margin":{"top":"8rem","bottom":"8rem"}}},"layout":{"inherit":true}} -->
+<main class="wp-block-group" style="margin-top:8rem;margin-bottom:8rem">
```
carolinan, line 4: I did not find a matching spacing preset for this margin.

theme.json
```diff
@@ -210,10 +210,22 @@
...
 					"fontSize": "var(--wp--preset--font-size--small)",
-					"fontStyle": "italic"
+					"fontWeight": "400"
+				},
+				"elements": {
+					"link": {
+						"typography": {
+							"textDecoration": "none"
+						}
+					}
```
carolinan: Removing underlines by default is not awesome but I think it is the least bad option.

theme.json
```diff
@@ -224,7 +236,16 @@
 			},
 			"core/post-title": {
 				"typography": {
-					"fontSize": "var(--wp--preset--font-size--xx-large)"
+					"fontSize": "var(--wp--preset--font-size--large)",
+					"letterSpacing": "-2px",
+					"fontWeight": "400"
+				},
```
carolinan: I believe I saw a discussion about changing the font weight site wide for the headings, but I don't recall where so I set it to 400 here to match the figma design.

templates/home.html
```diff
@@ -1,35 +1,44 @@
...
+	
+	<!-- wp:query {"queryId":0,"query":{"perPage":3,"pages":0,"offset":0,"postType":"post","order":"desc","orderBy":"date","author":"","search":"","exclude":[],"sticky":"","inherit":true},"displayLayout":{"type":"flex","columns":3},"align":"wide","layout":{"inherit":true}} -->
+	<div class="wp-block-query alignwide">
+		<!-- wp:group {"align":"wide","layout":{"type":"flex","flexWrap":"nowrap","justifyContent":"space-between"}} -->
+		<div class="wp-block-group alignwide">
+			<!-- wp:heading {"level":3,"style":{"typography":{"textTransform":"uppercase","fontStyle":"normal","fontWeight":"400"}},"fontSize":"small"} -->
+			<h3 class="has-small-font-size" style="font-style:normal;font-weight:400;text-transform:uppercase">Latests posts</h3>
+			<!-- /wp:heading -->
+			<!-- wp:query-pagination {"layout":{"type":"flex","justifyContent":"right"}} -->
+			<!-- wp:query-pagination-next {"label":"View more","style":{"typography":{"fontStyle":"normal","fontWeight":"400"}},"fontSize":"small"} /-->
```
carolinan: This font weight should be moved to theme.json.
The archive page query pagination has a different font size, and we need to decide which size the default should be.

carolinan: Home pagination font size: 18px, archive pagination font size: 16px. The theme does not have a size for 16 px?

templates/home.html
mikachan: I've just updated the spacing presets in another PR: https://github.com/WordPress/twentytwentythree/pull/52

I'm not sure if what I've done in that PR is what we should go with, but if we go down that route, there is a spacing preset that would fit `8rem` there.

templates/home.html
mikachan: > The theme does not have a size for 16 px?

I think the small font size should cover this - it's currently set to 1rem.

carolinan: I am finding it confusing because of the fluid typography. On the screen that I am using right now, the small font size is 22px. So I am unable to make it match the design without using pixels.

carolinan:
This is on Mac, chrome, at 400px.
The small font size is 12px.
The letter spacing looks bad:

Screen Shot 2022-08-16 at 04 23 40

carolinan:
Also I don't understand the letter spacing on figma. It says, on mobile, the letter spacing for H1 is 4rem?
https://www.figma.com/file/ohYEHkZzGjOse46NceTfj5/Twenty-Twenty-Three?node-id=301%3A708

beafialho:
Thank you, the template looks accurate.

@carolinan sorry some of Figma's specifications weren't right. I've updated them with a small change: all typography styles should have `0px` letter spacing.

Captura de ecrã 2022-08-16, às 10 05 56

templates/home.html
mikachan: Ah, I see. Using fluid typography does make it difficult to match the designs in all cases. I think it's OK to make a best guess, and @beafialho can advise on any further changes.

Maybe we could introduce some `min` and `max` values for the fluid font sizes, to constrain them?

carolinan: I think if the design allows using the same size, small, for both places that would solve it.

mikachan (approved):
I think this is ready to bring in 🎉

PR #52, opened 2022-08-15
Update spacing presets

### LINKED ISSUE #6: Match default spacing to Figma
The current spacing is using the new [spacing size presets](https://github.com/WordPress/gutenberg/pull/41527) from Gutenberg, however, the sizes are incorrect. They should match the spacing sizes used in Twenty Twenty-Two and the new Twenty Twenty-Three designs.

mikachan:
This PR updates the spacing presets and any spacing sizes that are in use to match TT2 more closely.

This is quite an awkward one to test so I'll summarise what I've tried to do. I've updated the spacing presets to try and cover TT2's spacing sizes like so:

- `var(--wp--preset--spacing--35)` should replace TT2's custom small size
- `var(--wp--preset--spacing--40)` should replace a common TT2 spacer size of 32px / 2rem
- `var(--wp--preset--spacing--45)` should replace TT2's custom medium size
- `var(--wp--preset--spacing--90)` should replace a common TT2 spacer size of 112px / 7rem
- `var(--wp--preset--spacing--85)` should replace TT2's custom large size

Closes https://github.com/WordPress/twentytwentythree/issues/6


theme.json
```diff
@@ -68,11 +68,28 @@
...
 			},
+			"spacingSizes": [
+				{
+					"size": "max(1.25rem, 5vw)",
+					"slug": "35",
+					"name": "Flexi Small"
+				},
+				{
+					"size": "clamp(2rem, 8vw, calc(4 * var(--wp--style--block-gap)))",
+					"slug": "45",
```
jffng: Is the `45` preset used anywhere? I only see `40` being used, so am curious why it's being added.

jffng (commented):
The main impact this PR has for me is adding more space between elements. I don't know if that's desirable or not, will leave to @beafialho to assess, but the spacing does appear to match TT2's spacing more closely. 

Trunk | This PR
--- | ---
Screen Shot 2022-08-15 at 3 13 44 PM | Screen Shot 2022-08-15 at 3 13 52 PM
 
This PR | TT2
--- | ---
Screen Shot 2022-08-15 at 3 13 52 PM | Screen Shot 2022-08-15 at 3 23 42 PM

Side note, something happened recently to the type sizes, I'm not sure why they all look a lot bigger.

theme.json
mikachan: No it's not currently being used anywhere, but this size was available with TT2 so I thought it would be good to include it, especially as it's a fluid preset.

beafialho:
> The main impact this PR has for me is adding more space between elements. I don't know if that's desirable or not, will leave to @beafialho to assess, but the spacing does appear to match TT2's spacing more closely.

I can't test this but from @jffng's screenshots I agree, it looks closer to TT2 👍 

The spacing between the post title and excerpt seems a bit too much to me but I'd only be able to tell if I tested it.

theme.json
richtabor: What's the reasoning behind using the `X5` instead of overriding core's `X0` values? (i.e. `35` instead of `30`).

mikachan: The reason I went with `X5` values was because they're fluid values, and this set them apart from the `X0` values.

As they're fluid, they don't necessarily fit between the other sizes, as these value names suggest. But I'm also not sure what else to call them!

pbking: Could the X0 values be fluid instead?

mikachan: I think we'd need to manually specify each preset in the `spacingSizes` array to make each size fluid. Or is there currently a way to make each size in `spacingScale` fluid automatically?

Or maybe it would be enough to specify each TT2 fluid size (small, medium, and large) in the `spacingSizes` array as fluid presets, and not have any added by `spacingScale`?

mikachan:
After playing around with some different spacing preset variants, I've updated this PR with the following:

- Removed the `spacingScale` setting, so there are no spacing presets added automatically
- Added the 5 required sizes mentioned in the PR description as fluid presets in the `spacingSizes` array

This means that these are now the presets:

- `var(--wp--preset--spacing--30)` replaces TT2's custom fluid small size
- `var(--wp--preset--spacing--40)` replaces a common TT2 spacer size of 32px / 2rem, as a fluid value
- `var(--wp--preset--spacing--50)` replaces TT2's custom fluid medium size
- `var(--wp--preset--spacing--60)` replaces TT2's custom fluid large size
- `var(--wp--preset--spacing--70)` replaces a common TT2 spacer size of 112px / 7rem, as a fluid value

I really like how this turned out, I think the biggest downside is that we're no longer showcasing how the `spacingScale` setting can work. But maybe we could add this to a style variation... Plus we're now showcasing the `spacingSizes` array instead.

carolinan:
How do these relate to the design rather than to TT2?
The templates need to be updated. Should `spacing 70` be used even if the design uses more than 112px? For example, on the homepage the distance between the image and the heading is 129px.

mikachan:
Yeah this is a good point, I think we should decide which presets match the spacing sizes in the design, and if these sizes can't cover all possibilities, then add more.

beafialho:
I'm not sure if this is helpful at this point, but I've noticed the spacings on Figma weren't pixel perfect, so I've updated them anyway and it should be more thorough now.

You can see both the spacers and the grid:

https://user-images.githubusercontent.com/65220155/185364866-5e193692-6b51-4744-852f-25a2b5bcf200.mp4

carolinan:
I solved the merge conflicts. The search template still has hardcoded rems in it that should be updated.

mikachan:
I think this is ready for another review. I've simplified some of existing presets, and added an additional `var(--wp--preset--spacing--80)` size. 

Here's a table of spacing size presets, with the other sizes for reference:

| TT2 Equivalent  | TT3 Design | TT3 Actual | TT3 Variable |
| ----------- | --------- | ----------- | ------------ |
| custom-small | 1.5rem - 2.125rem | 1.5rem - 2rem |  var(--wp--preset--spacing--30) |
| 32px / 2rem | 1.875rem - 3.125rem | 1.8rem - 3rem |  var(--wp--preset--spacing--40) |
| custom-medium | 2.438rem - 6.5rem | 2.5rem - 6.5rem |  var(--wp--preset--spacing--50) |
| custom-large | 3.75rem - 7.125rem | 3.75rem - 7rem |  var(--wp--preset--spacing--60) |
| 112px / 7rem | 5.125rem - 8rem | 5rem - 8rem |  var(--wp--preset--spacing--70) |
| n/a | 7.125rem - 11rem | 7rem - 11rem |  var(--wp--preset--spacing--80) |

I think the TT2 spacing should be the least important comparison, but I found it useful to include in order to compare any spacing values copied over from the TT2 templates (although, there are not many left now.)

The TT3 design spacing is based on the [designs here](https://www.figma.com/file/j469TEkvf2YJ3EVQm9X4Nt/TT3?node-id=147%3A1745), where I saw the following sizes:

- 24px / 1.5rem
- 30px / 1.875rem
- 34px / 2.125rem
- 39px / 2.438rem
- 44px / 2.75rem
- 50px / 3.125rem
- 60px / 3.75rem
- 82px / 5.125rem
- 104px / 6.5rem
- 114px / 7.125rem
- 129px / 8rem
- 176px / 11rem

They can't be mapped 1:1 as we're using fluid spacing, but I think we're covering most sizes with the above table.

templates/archive.html
```diff
@@ -1,21 +1,21 @@
...
-<main class="wp-block-group" style="margin-top:8rem;margin-bottom:8rem">
+<!-- wp:group {"tagName":"main","style":{"spacing":{"margin":{"top":"var(--preset--spacing--70)","bottom":"var(--preset--spacing--70)"}}},"layout":{"inherit":true}} -->
+<main class="wp-block-group" style="margin-top:var(--preset--spacing--70);margin-bottom:var(--preset--spacing--70)">
 	<!-- wp:query-title {"type":"archive","align":"wide","style":{"spacing":{"margin":{"bottom":"6.4rem"}}}} /-->
 	<!-- wp:query {"query":{"perPage":6,"pages":0,"offset":0,"postType":"post","order":"desc","orderBy":"date","author":"","search":"","exclude":[],"sticky":"","inherit":true},"displayLayout":{"type":"flex","columns":3},"align":"wide","layout":{"inherit":false}} -->
 	<div class="wp-block-query alignwide">
 		<!-- wp:post-template {"align":"wide"} -->
 			<!-- wp:post-featured-image {"isLink":true,"align":"wide"} /-->
-			<!-- wp:post-title {"isLink":true,"style":{"spacing":{"margin":{"top":"var(--preset--spacing--60)","bottom":"var(--preset--spacing--60)"}}}} /-->
+			<!-- wp:post-title {"isLink":true,"style":{"spacing":{"margin":{"top":"var(--preset--spacing--40)","bottom":"var(--preset--spacing--40)"}}}} /-->
```
jffng: I think these are missing the `--wp` prefix, also it made me think — should we apply this same spacing to the Latest Posts section on the Home template? 

```suggestion
			<!-- wp:post-title {"isLink":true,"style":{"spacing":{"margin":{"top":"var(--wp--preset--spacing--40)","bottom":"var(--wp--preset--spacing--40)"}}}} /-->
```

templates/home.html
```diff
@@ -28,9 +28,9 @@ <h3 class="has-small-font-size" style="font-style:normal;font-weight:400;text-tr
 			<!-- wp:post-title {"level":4,"isLink":true,"style":{"spacing":{"margin":{"top":"var(--wp--preset--spacing--60)","bottom":"var(--wp--preset--spacing--60)"}}}} /-->
```
jffng: Should these be `40` or something smaller to match the design more closely?

```suggestion
			<!-- wp:post-title {"level":4,"isLink":true,"style":{"spacing":{"margin":{"top":"var(--wp--preset--spacing--40)","bottom":"var(--wp--preset--spacing--40)"}}}} /-->
```

templates/archive.html
mikachan: Ugh, yes, good spot!

> should we apply this same spacing to the Latest Posts section on the Home template?

Yeah, I think that would work 👍

templates/home.html
mikachan: I think 40 looks better in this case.

jffng (approved):

carolinan:
> 
> Here's a table of spacing size presets, with the other sizes for reference:
> 

I keep coming back to this list, could this be added to the readme or any other type of easy to access documentation?
It would be good to have such a reference for the font sizes too.

mikachan:
Yeah this is a great idea, I also keep opening this PR just to reference this table. I'll spin up a PR for some additional docs...

PR #53, opened 2022-08-16
Remove default italics

carolinan:
Some blocks had an italic style as default, this does not match the design.


theme.json
```diff
@@ -212,14 +211,12 @@
...
-					"fontSize": "var(--wp--preset--font-size--small)",
-					"fontStyle": "italic"
+					"fontSize": "var(--wp--preset--font-size--small)"
 				}
 			},
 			"core/post-terms": {
 				"typography": {
-					"fontSize": "var(--wp--preset--font-size--small)",
-					"fontStyle": "italic"
+					"fontSize": "var(--wp--preset--font-size--small)"
```
carolinan, line 272: I found that the post tags are italic, but not the categories.
We should avoid adding italics to the markup, because we can't remove that with style variations. So for now, the italic is removed. A decision is needed.

mikachan: > We should avoid adding italics to the markup, because we can't remove that with style variations.

This sounds good to me, I think it's good to have a general rule that settings should be controlled by style variations where possible. Does this sound OK to you, @beafialho?

beafialho: Yes, this sounds ok to me.

mikachan (approved):

PR #54, opened 2022-08-16
Move JSON schema and version prop to top

thelovekesh:
Move `version` and `$schema` to the top of `theme.json` file. 

Reasons:
- The version property should be visible as soon as someone opens the theme.json
- After getting many references, I found that defining `$schema` at the file top make much more sense. Ref: https://json-schema.org/learn/getting-started-step-by-step.html

### DIFF
--- a/theme.json
+++ b/theme.json
@@ -1,4 +1,6 @@
 {
+	"$schema": "https://schemas.wp.org/trunk/theme.json",
+	"version": 2,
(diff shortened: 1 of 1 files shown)


thelovekesh:
cc: @mikachan

Soean (approved):

mikachan (approved):

PR #56, opened 2022-08-16
Fix pattern texts

carolinan:
The translatable text in the patterns needs to be output, not returned, or it will not be displayed.

### DIFF
--- a/patterns/hidden-404.php
+++ b/patterns/hidden-404.php
@@ -6,9 +6,9 @@
  */
 ?>
 <!-- wp:heading {"className":"has-text-align-center"} -->
-<h2 class="has-text-align-center"><?php esc_html( _x( '404', 'Error code for a webpage that is not found.', 'twentytwentythree' ) ); ?></h2>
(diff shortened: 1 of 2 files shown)


patterns/hidden-404.php
```diff
@@ -6,9 +6,9 @@
  */
 ?>
 <!-- wp:heading {"className":"has-text-align-center"} -->
-<h2 class="has-text-align-center"><?php esc_html( _x( '404', 'Error code for a webpage that is not found.', 'twentytwentythree' ) ); ?></h2>
+<h2 class="has-text-align-center"><?php echo esc_html( _x( '404', 'Error code for a webpage that is not found.', 'twentytwentythree' ) ); ?></h2>
```
jffng: What is the reason for using `echo esc_html` here and `esc_html_e` below?

jffng (approved):
Thank you, this fixes the 404 and footer output.

patterns/hidden-404.php
carolinan: Hmm actually it should be `echo esc_html_x()`:  [esc_html_x](https://developer.wordpress.org/reference/functions/esc_html_x/) does not have variant that both echoes, escapes, translates, and uses the context

PR #57, opened 2022-08-16
Add alternative-blog.html

carolinan:
Adds the alternative blog template from the design.

### DIFF
--- a/templates/alternative-blog.html
+++ b/templates/alternative-blog.html
@@ -1,0 +1,22 @@
+<!-- wp:template-part {"slug":"header","tagName":"header"} /-->
+
+<!-- wp:group {"tagName":"main","layout":{"inherit":true}} -->
+<main class="wp-block-group">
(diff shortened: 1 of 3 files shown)


mikachan (commented):
Thanks for this, it looks good to me ✨

beafialho:
~~This isn't looking like the designs for me: it's missing the separators, the post date isn't with the same format and letter case.~~

My bad, I wasn't testing correctly. My only comment is that the query loop should be wide instead.

carolinan:
I will change it to wide.

-What do you think the name of the file and the visual label in the editor should be?

Assuming this should be selectable in the drop down for pages, for example when the "Front page displays..." setting is used to select a separate page for the blog.

beafialho:
> -What do you think the name of the file and the visual label in the editor should be?

I think it could be **Blog (Alternative)**.

mikachan (approved):

PR #61, opened 2022-08-16
Add the home page "call to action" / signup pattern

carolinan:
Adding this pattern here for now, and we can discuss if it should be in the pattern directory or the theme.

I am concerned the text might indicate to some users that the theme has some sort of newsletter functionality.
Can we come up with something more neutral?

### DIFF
--- a/patterns/call-to-action.php
+++ b/patterns/call-to-action.php
@@ -9,17 +9,16 @@
 ?>
 <!-- wp:columns {"align":"wide"} -->
 <div class="wp-block-columns alignwide">
-	<!-- wp:column {"width":"40%"} -->
-	<div class="wp-block-column" style="flex-basis:40%">
-		<!-- wp:paragraph {"fontSize":"x-large"} -->
-		<p class="has-x-large-font-size">
-		<?php esc_html_e( 'Get daily reflections in your inbox.', 'twentytwentythree' ); ?>
-		</p><!-- /wp:paragraph -->
+	<!-- wp:column -->
+	<div class="wp-block-column">
+		<!-- wp:paragraph {"style":{"typography":{"lineHeight":"1.2"}},"fontSize":"x-large"} -->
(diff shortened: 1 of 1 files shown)


beafialho:
> I am concerned the text might indicate to some users that the theme has some sort of newsletter functionality.
> Can we come up with something more neutral?

Good point, how about this?

Captura de ecrã 2022-08-17, às 10 44 02

I noticed the line height seems a little too large to me, compared to the Figma specs:

Captura de ecrã 2022-08-17, às 10 47 52

carolinan:
When I updated the text, the question mark was on a third line. So I removed the 40% width of the left column.
That meant that the line did not extend as close to the text as in the design:

Screen Shot 2022-08-17 at 12 10 52

We should probably test with a row block too instead of a columns block.

carolinan:
I tested a row with a stack with the paragraph and button on the left, and the separator on the right, and it did not work, the separator is invisible.

row block with nested blocks

beafialho:
> When I updated the text, the question mark was on a third line. So I removed the 40% width of the left column.
> That meant that the line did not extend as close to the text as in the design:

Thank you, I think that's ok.

patterns/call-to-action.php
```diff
@@ -0,0 +1,34 @@
...
+ * Keywords: Call to action, signup, newsletter
+ * Block Types: core/buttons
+ */
+?>
+<!-- wp:columns {"align":"wide"} -->
+<div class="wp-block-columns alignwide">
+	<!-- wp:column -->
+	<div class="wp-block-column">
+		<!-- wp:paragraph {"style":{"typography":{"lineHeight":"1.2"}},"fontSize":"x-large"} -->
+		<p class="has-x-large-font-size" style="line-height:1.2">Got any book recommendations?</p>
```
mikachan: Does this content need to be escaped/translated?

carolinan: Yes!

mikachan (approved):
🚢🚢

PR #63, opened 2022-08-16
Update 404 page pattern desgin

### LINKED ISSUE #62: Update the font sizes and spacing on the 404 page.
The 404 page pattern needs to be updated to match [the design](https://www.figma.com/file/ohYEHkZzGjOse46NceTfj5/Twenty-Twenty-Three?node-id=301%3A522).

tanvirulhaque:
The 404 page pattern is updated to match the [design](https://www.figma.com/file/ohYEHkZzGjOse46NceTfj5/Twenty-Twenty-Three?node-id=301%3A522).


patterns/hidden-404.php
```diff
@@ -5,10 +5,17 @@
...
-<p class="has-text-align-center"><?php esc_html_e( 'This page could not be found. Maybe try a search?', 'twentytwentythree' ); ?></p>
+
+<!-- wp:paragraph {"align":"center","fontSize":"small"} -->
+<p class="has-text-align-center"><?php esc_html_e( 'This page could not be found.', 'twentytwentythree' ); ?></p>
 <!-- /wp:paragraph -->
-<!-- wp:search {"label":"Search","showLabel":false,"width":50,"widthUnit":"%","buttonText":"Search","buttonUseIcon":true,"align":"center"} /-->
+
+<!-- wp:spacer {"height":"var(--wp--preset--spacing--100)"} -->
+<div style="height:var(--wp--preset--spacing--100)" class="wp-block-spacer"></div>
+<!-- /wp:spacer -->
```
mikachan, line 19: The spacer block is showing an invalid content error for me:

image


If I replace the markup with the below, it seems to fix the error:

```
<!-- wp:spacer {"height":"var(--wp--preset--spacing--100)"} -->
<div style="height:var(--wp--preset--spacing--100)" aria-hidden="true" class="wp-block-spacer"></div>
<!-- /wp:spacer -->
```

mikachan (commented):
This is looking great, thanks for working on it! I can see it matches the design well. I've left one small inline comment.

I do have a comment about how we need to set the font size, line height, and font weight for the `h2` directly in the pattern. It would be great if we could move each of these settings out of the pattern and into the theme.json file. This would give style variations more control over this pattern.

I know this is a unique size that is unlikely to be used anywhere else, so perhaps we should leave the settings in the pattern in this case. Also happy for this to happen in a separate PR!

patterns/hidden-404.php
tanvirulhaque: Ok, I have added the `aria-hidden="true"` now invalid content error notice is not showing but the gap is not showing for me in the block editor.

mikachan: Hmm, I see this too. I have seen this recently elsewhere, although I thought it had been fixed or it was a problem with my environment. I think we should open an upstream issue for this in [Gutenberg](https://github.com/WordPress/gutenberg/issues). Would you like to handle this? Or I can do it, if you prefer.

Otherwise, I think this PR is good to bring in 🎉

mikachan (approved):

patterns/hidden-404.php
tanvirulhaque: Ok, you can open an upstream issue for this in Gutenberg.

carolinan: Do you have Gutenberg enabled?
If you create this block using WordPress 6.0.1 only, the aria-hidden is not included and the spacing presets do not work.

mikachan: I can see the same as @carolinan now, it seems I wasn't on Gutenberg trunk when I commented earlier (I thought I'd seen it working 😄). Sorry for the confusion! This is working in the editor:

image

Sounds like you probably need to activate Gutenberg if you also weren't seeing the aria-hidden attribute (and I'd recommend pulling the latest version from trunk).

tanvirulhaque: Sorry for the confusion @mikachan unfortunetly I was similar as yours. 😄

By the way!

> I know this is a unique size that is unlikely to be used anywhere else, so perhaps we should leave the settings in the pattern in this case. Also happy for this to happen in a separate PR!

I also agree with you.

PR #65, opened 2022-08-16
Add IBM Plex Mono to fonts collection

### LINKED ISSUE #12: New font: IBM Plex Mono

pbking:
Fixes #12 

This adds IBM Plex Mono to the collection of fonts that can be used.

This is an alternative to both #17 which was missing a weight and had it in a variation and #23 which also added it exclusively to a variation as well as other adjacent changes.

This change uses the fonts resources from #23 and adds them to the base theme.json file.

It's worth noting that the weights that are ultimately included are probably subject to change as the final released files should be carefully considered.  However this at least makes the font family useable today.


mikachan (approved):
This is brilliant, thank you!

PR #78, opened 2022-08-18
Update search.html

### LINKED ISSUE #25: Add the no results block
With the no results block, the theme can conditionally display blocks when a query returns no results.
Without this block, a search that has no results for example, displays an empty page.

The block requires WordPress 6.0 and depends on https://github.com/WordPress/twentytwentythree/issues/24

### LINKED ISSUE #77: Search template
Now that this should be possible ([#42662](https://github.com/WordPress/gutenberg/pull/42662)), we should match the Search template to the [designs](https://www.figma.com/file/ohYEHkZzGjOse46NceTfj5/Twenty-Twenty-Three?node-id=428%3A410).

Search

carolinan:
Updates the search template.
Closes https://github.com/WordPress/twentytwentythree/issues/77

localhost_10003_page_2__s=block

No results:
localhost_10003__s=books

### DIFF
--- a/templates/search.html
+++ b/templates/search.html
@@ -26,7 +26,7 @@
 		<p>Sorry, but nothing matched your search terms. Please try again with some different keywords.</p>
 		<!-- /wp:paragraph -->
 
-		<!-- wp:search {"label":"Search","showLabel":false,"buttonText":"Search"} /-->
+		<!-- wp:search {"label":"Search","showLabel":false,"buttonText":"Search","buttonUseIcon":true} /-->
 		<!-- /wp:query-no-results -->
 	</div>
 	<!-- /wp:query -->
--- a/templates/search.html
+++ b/templates/search.html
@@ -1,38 +1,35 @@
 <!-- wp:template-part {"slug":"header","tagName":"header"} /-->
 
-<!-- wp:group {"layout":{"inherit":true}} -->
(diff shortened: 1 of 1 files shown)


carolinan:
We can move the no results above or below the separator, but I don't know a way to hide it completely.
_Or, perhaps a top border could be added to the query pagination instead. Untested._

templates/search.html
```diff
@@ -1,38 +1,35 @@
...
+		<!-- wp:query-pagination {"paginationArrow":"arrow","layout":{"type":"flex","justifyContent":"space-between"}} -->
+		<!-- wp:query-pagination-previous {"label":"Newer Posts"} /-->
+		<!-- wp:query-pagination-next {"label":"Older Posts"} /-->
+		<!-- /wp:query-pagination -->
+		<!-- wp:query-no-results -->
+		<!-- wp:paragraph -->
+		<p>Sorry, but nothing matched your search terms. Please try again with some different keywords.</p>
+		<!-- /wp:paragraph -->
+
+		<!-- wp:search {"label":"Search","showLabel":false,"buttonText":"Search"} /-->
```
mikachan: Should we use the icon in the button here, like the 404 template?

mikachan:
I've just tried adding a border to the query pagination. It seems to work in the editor but not on the front end. This is the markup I tested with:

```
<!-- wp:query-pagination {"style":{"border":{"top":{"width":"2px"}},"spacing":{"padding":{"top":"var(--wp--preset--spacing--70)"}}},"paginationArrow":"arrow","layout":{"type":"flex","justifyContent":"space-between"}} -->
```

beafialho (approved):

PR #80, opened 2022-08-18
Replace default link color variable

mikachan:
The default link color variable is currently set to `foreground`, which isn't used in this theme. This PR replaces it with `contrast`.


pbking (approved):
👍 LGTM

PR #82, opened 2022-08-19
Add minimum value for small and medium sizes

### LINKED ISSUE #76: The 'small' font size scales down too much at smaller resolutions
Currently, the 'small' font size is set to `1rem`, but as we're using fluid typography and we have set no `min` or `max` value, this font size preset scales to a very small size at smaller resolutions (12px).

I think we should add at least a `min` value to the 'small' size, maybe around `14px` / `0.9rem`?

mikachan:
This PR adds a `min` value for the small and medium fluid font sizes, as I believe they are scaling down too much at smaller resolutions.

- Small: was scaling down to `12px`, now fixed at a minimum value of `14px` / `0.875rem`
- Medium: was scaling down to `13.5px`, now fixed at a minimum value of `16px` / `1rem`

Closes https://github.com/WordPress/twentytwentythree/issues/76.

### DIFF
--- a/theme.json
+++ b/theme.json
@@ -219,14 +219,14 @@
 			"fontSizes": [
 				{
 					"fluid": {
-						"min": "14px"
+						"min": "0.875rem"
 					},
 					"size": "1rem",
 					"slug": "small"
 				},
 				{
 					"fluid": {
-						"min": "16px"
+						"min": "1rem"
 					},
 					"size": "1.125rem",
 					"slug": "medium"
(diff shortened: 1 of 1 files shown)


theme.json
```diff
@@ -218,10 +218,16 @@
...
 				{
+					"fluid": {
+						"min": "14px"
+					},
 					"size": "1rem",
 					"slug": "small"
 				},
 				{
+					"fluid": {
+						"min": "16px"
```
MaggieCabrera: should the units be rems?

mikachan: Yeah this would be better! Not sure why I defaulted to px 😄

MaggieCabrera (approved):
👍

PR #83, opened 2022-08-19
Update navigation and site title styles

### LINKED ISSUE #74: Apply the navigation block styles to the sub menu block and home link
I noticed that the font size and hover/focus/active is different on the submenu and home link compared to the "core/navigation-link", which is styled in theme.json.

Is it possible to place these styles directly on the navigation block instead of the child block?
If not, I think that if we are going to style one of them, we have to style and test all the child blocks.

mikachan:
This PR moves the link styles from the child `navigation-link` block to the parent `navigation` block.

It also replicates these link styles on the `site-title` block, so that the site title link states match the navigation link states.

Closes https://github.com/WordPress/twentytwentythree/issues/74.

### DIFF
--- a/theme.json
+++ b/theme.json
@@ -449,13 +449,21 @@
 				},
 				":hover": {
 					"typography": {
-						"textDecoration": "underline dashed"
+						"textDecoration": "none"
 					}
 				},
 				":active": {
 					"typography": {
 						"textDecoration": "none"
 					}
+				},
+				":focus": {
+					"typography": {
+						"textDecoration": "none"
+					}
+				},
+				"typography": {
+					"textDecoration": "underline"
 				}
 			}
 		},
--- a/theme.json
+++ b/theme.json
@@ -312,6 +312,21 @@
 				},
 				"elements": {
 					"link": {
+						":hover": {
+							"typography": {
+								"textDecoration": "underline"
+							}
+						},
+						":focus": {
+							"typography": {
+								"textDecoration": "underline"
+							}
+						},
+						":active": {
+							"typography": {
+								"textDecoration": "underline"
+							}
+						},
 						"typography": {
 							"textDecoration": "none"
 						}
@@ -452,12 +467,12 @@
 						"textDecoration": "none"
 					}
 				},
-				":active": {
+				":focus": {
 					"typography": {
 						"textDecoration": "none"
 					}
 				},
-				":focus": {
+				":active": {
 					"typography": {
 						"textDecoration": "none"
 					}
--- a/theme.json
+++ b/theme.json
@@ -342,6 +342,23 @@
 				}
 			},
 			"core/site-title": {
+				"elements": {
+					"link": {
+						":hover": {
+							"typography": {
+								"textDecoration": "underline"
+							}
+						},
+						":focus": {
+							"typography": {
+								"textDecoration": "underline"
+							}
+						},
+						"typography": {
+							"textDecoration": "none"
+						}
+					}
+				},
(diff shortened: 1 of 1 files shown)


theme.json
```diff
@@ -342,6 +342,23 @@
 				}
 			},
 			"core/site-title": {
+				"elements": {
+					"link": {
+						":hover": {
+							"typography": {
+								"textDecoration": "underline"
```
carolinan, line 376: Should the underline be dashed or solid? 🤔

carolinan:
In post content:
The page list block, home link, site title and the submenu block have the green link color on the front, but not in the editor where they are black.
The navigation link and page link are black in both views.

All text links that can be added to the navigation block are underlined by default in the editor, except the site title block. -This is different from the front where the underline is hidden in the default state.

If I turn on "open on click":
Parent items that have sub menus are not underlined on hover anymore -they are buttons. As a *user* I would expect the menu to look the same when I enable this option.
Parent items in the page list block has a black text color but the sub menu item links are green.

beafialho:
Some things I noticed:

- Site title and nav links hover states are looking good.
- The reverse should happen when we hover any underlined link by default (paragraph link, categories, tags, log in to comment, credit in the footer and so on).

https://user-images.githubusercontent.com/65220155/185913934-fb266895-4534-456c-997f-065a672a3ee6.mp4

- Post titles currently have no underline by default but they should show an underline when we hover them.

https://user-images.githubusercontent.com/65220155/185914835-322c0179-3de4-488f-866b-9f13bf3c8fc8.mp4

> The navigation link and page link are black in both views.

Those are supposed to be black. Only links in paragraphs are intended be green.

theme.json
mikachan: @beafialho Could you confirm which underlines should be dashed and which should be solid?

mikachan:
Thanks for testing @carolinan!

> The page list block, home link, site title and the submenu block have the green link color on the front, but not in the editor where they are black.

I can't see this in my local editor. I'll keep an eye on this and check for any related GB issues.

> All text links that can be added to the navigation block are underlined by default in the editor, except the site title block. -This is different from the front where the underline is hidden in the default state.

I can see this. I think I've fixed it by removing underlines by default on the navigation block.

> Parent items that have sub menus are not underlined on hover anymore -they are buttons. As a user I would expect the menu to look the same when I enable this option.

I see this too - I can't seem to fix it with additional settings in theme.json, I'll find/open an upstream issue.

> Parent items in the page list block has a black text color but the sub menu item links are green.

I can't see this either, this sounds similar to the first issue you listed. It could totally be my environment!

theme.json
beafialho: Does this help @mikachan?

Captura de ecrã 2022-08-22, às 15 29 46

mikachan:
Thanks @beafialho!

> The reverse should happen when we hover any underlined link by default (paragraph link, categories, tags, log in to comment, credit in the footer and so on).

This should now be the case with every other block other than navigation and navigation-link.

> Post titles currently have no underline by default but they should show an underline when we hover them.

This should be done.

beafialho:
Thanks @mikachan, I see those changes now!

theme.json
mikachan: Yes, thank you!

mikachan:
I've updated this PR so the links should now match the designs from [here](https://github.com/WordPress/twentytwentythree/pull/83#discussion_r951514268).

carolinan (approved):

Issue #84, opened 2022-08-19
Remove opinionated settings from block markup


mikachan:
As TT3 is focused on style variations, it would be good to remove as many opinionated styles from the block markup as possible. There are a few places where this is happening, but sometimes it can be difficult where unique styles have been required. Here's an example of where we have done this previously: https://github.com/WordPress/twentytwentythree/pull/31

This issue is to track where we could consider moving these settings, and encourage discussion around this for each area:

- [x] 404 pattern: https://github.com/WordPress/twentytwentythree/pull/93
- [x] Search template: https://github.com/WordPress/twentytwentythree/pull/103
- [x] Home template: https://github.com/WordPress/twentytwentythree/pull/94

beafialho:
I agree 404 can become much smaller:

<img width="1148" alt="Captura de ecrã 2022-08-22, às 09 53 33" src="https://user-images.githubusercontent.com/65220155/185881213-2f536bc0-d302-4cfe-9c27-1ef55eaf6e93.png">

About Search and Home, I'm not sure. What changes do you think could be done to make it less opinionated?

pbking:
I don't see that the xx-large size has been used anywhere yet.  Would it make sense to use that value, but maybe crank it up a bit? I'm actually keen to keep the value on the 404 page (`font-size:clamp(4rem, 40vw, 20rem)` or something close to it) for xx-large (which feels like "the upper end of sizes we should create").

pbking:
I didn't notice any opinion in the search template that we might want to relocate (other than some spacing bits that are already being addressed in #52

mikachan:
I'm going to close this, as I think we've currently removed as much opinion as possible from the templates. Please open another issue if you spot anything similar.

PR #93, opened 2022-08-23
Moved 404 size opinion to theme.json

pbking:
This change moves the font size of the large "404" to theme.json (as the xx-large font size which was currently otherwise unused).  This is a 'fluid' size.

It also fixes a small block markup error in the paragraph block that follows it.

Result:
Small:
image

Medium:
image

Large:
image

Addresses some of #84


mikachan (approved):
Thanks for this! LGTM!

PR #94, opened 2022-08-23
Moved sizing opinions from home to theme.json

pbking:
This change increases the font weight of H2 and H3 to 400.  The only places I've noticed it's used is in the home template and both of those instances were increasing the font weight so I suggest we do that in theme.json instead.

The 'Latest Posts' text was an H3 customized to nearly the same values assigned to H6 so i just changed it to an H6.

Before:

image

image

After:
image

image

Partially addresses #84

### DIFF
--- a/templates/home.html
+++ b/templates/home.html
@@ -6,18 +6,18 @@
 	<!-- wp:image {"align":"wide"} -->
 	<figure class="wp-block-image alignwide"><img alt=""/></figure>
 	<!-- /wp:image -->
-	
-	<!-- wp:heading {"align":"wide","style":{"typography":{"fontWeight":"400"},"spacing":{"margin":{"top":"8rem","bottom":"8rem"}}}} -->
-	<h2 class="alignwide" style="margin-top:8rem;margin-bottom:8rem;font-weight:400;">Mindblown: a blog about philosophy.</h2>
+	<!-- wp:heading {"align":"wide","style":{"spacing":{"margin":{"top":"8rem","bottom":"8rem"}}}} -->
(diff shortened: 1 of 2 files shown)


carolinan:
Please never change heading levels because you need to match a style :)
Using the correct heading levels are important for those who navigate via heading levels, like some screen reader users.

mikachan (commented):
The font-weight changes are great, the more we can move out of the markup, the better!

In regards to the Latest Posts heading, I agree with @carolinan that it should remain a `h3`, but then this makes the styling tricky. It looks like we either need to keep the styling in the markup for this or change the `h3` styles. Or perhaps we could remove this heading, as suggested [here](https://github.com/WordPress/twentytwentythree/issues/43#issuecomment-1222063777)?

carolinan:
Yeah I agree with removing the texts that are above the posts, and placing the query pagination and the separator below.

madhusudhand:
I see one case where the sizing needs to be added in the markup.
For **post title**, even though it is given `h4`, the `font-size` is derived from `core/post-title` because of the higher specificity.

```
<!-- wp:post-title {"level":4,"isLink":true,"style":{"spacing":{"margin":{"top":"var(--wp--preset--spacing--60)","bottom":"var(--wp--preset--spacing--60)"}}}} /-->
```

This needs fontSize defined in the template to match the figma.

| Current | Expected |
| --- | --- |
| image | image |

**Update:** #109 fixes the issue.

mikachan:
I completely forgot this PR was ongoing when I created this one: https://github.com/WordPress/twentytwentythree/pull/111

Happy for this one to be preferred!

pbking:
Nah, I'll close this one in deference to #111 and add any comments to that one.

PR #99, opened 2022-08-23
Add the call to action block pattern to the home template

### LINKED ISSUE #43: Home template: match the Figma

carolinan:
Adds the call to action block pattern to the home template, inside the main element.

I also tried an alternative where the pattern was inside a separate `<section>` between the `<main>` and the `<footer>`.
I switched to including it in the main element because I believe that if a section is used, it will need an aria-label.
I don't think `<aside>` would have been a good option either since it's a call to action.

In the pattern, I needed to place the opening `<?php` on the same line as the p tag, because there was an extra space before the first letter.


mikachan (approved):
This looks good to me. I agree with your reasoning around placing this in the `main` tag 👍

Issue #95, opened 2022-08-23
Spacing: replace var() format with var:preset|spacing|value


carolinan:
Across the templates, the format `var(--wp--preset--spacing--90)` is used inside the JSON block comment.
For the code in the template to match the code of new blocks added in the block editor, this format should be used: `var:preset|spacing|numeric-value`
See https://github.com/WordPress/twentytwentythree/pull/55#discussion_r952127903

carolinan:
Updating it for the spacer block did not work:
```
<!-- wp:spacer {"height":"var:preset|spacing|40"} -->
<div style="height:var(--wp--preset--spacing--40)" aria-hidden="true" class="wp-block-spacer"></div>
<!-- /wp:spacer -->
```

```
Block validation failed for `core/spacer` (Object).
Content generated by `save` function:
<div style="height:var:preset|spacing|40" aria-hidden="true" class="wp-block-spacer"></div>
Content retrieved from post body:
<div style="height:var(--wp--preset--spacing--40)" aria-hidden="true" class="wp-block-spacer"></div>
```

carolinan:
Updating it for the post title did not work:
`<!-- wp:post-title {"level":1,"align":"wide","style":{"spacing":{"margin":{"top":"var:preset|spacing|40","bottom":var:preset|spacing|40"}}}} /-->`
-Shows no margin in the site editor.

mikachan:
It looks like we've replaced all the possible spacing presets with the correct format across the templates. 

The only ones left are the spacer blocks, but it looks like this currently isn't possible as the spacing presets are not valid height settings. I've opened up an issue for this: https://github.com/WordPress/gutenberg/issues/43816

PR #103, opened 2022-08-23
Search and archive templates: update spacing

mikachan:
This PR includes a couple of changes, mainly adjusting the spacing variables in the search template:

- fixes a typo with the `spacing--50` preset
- replaces the spacing variables with this style: `var:preset|spacing|70` in the JSON markup (instead of this: `var(--wp--preset--spacing--70)`) (part of https://github.com/WordPress/twentytwentythree/issues/95, and mentioned here: https://github.com/WordPress/twentytwentythree/pull/55#discussion_r952127903)
- replaces a `spacing--60` with a `spacing--40`, as I thought the spacing looked slightly too big here

### DIFF
--- a/theme.json
+++ b/theme.json
@@ -76,7 +76,7 @@
 					"name": "40"
 				},
 				{
-					"size": "clamp(2.5rem, 8vw, 6.5rem))",
+					"size": "clamp(2.5rem, 8vw, 6.5rem)",
 					"slug": "50",
 					"name": "50"
 				},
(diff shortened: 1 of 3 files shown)


theme.json
```diff
@@ -76,7 +76,7 @@
 					"name": "40"
 				},
 				{
-					"size": "clamp(2.5rem, 8vw, 6.5rem))",
+					"size": "clamp(2.5rem, 8vw, 6.5rem)",
```
madhusudhand, line 79: Sizes `40` and `70` are using the calculation from [here](https://websemantics.uk/tools/responsive-font-calculator/), where as others not. Use the same calculation for all?

mikachan: > Use the same calculation for all?

Yes, we could do this. The main reason I didn't opt to use this for all of the spacing presets was because I was trying to stay close to the TT2 implementation - the ones that use the VW calculation are new sizes. It took a while to wrap up the initial spacing PR too 😅 But now that we have the initial version merged, it would be great to revisit with improvements!

carolinan:
The archive template should match this.

mikachan:
> The archive template should match this.

Yes, thank you! This should now be done.

carolinan (approved):

PR #109, opened 2022-08-24
fix fluid typography and block font sizes

madhusudhand:
This addresses the issues of fluid typography and font-sizes of few blocks.

* Add `max` for `small` and `medium` to avoid max being set to a factor of 1.5 by default by Gutenberg.
* Add an explicit `fluid: false` for `large` and `x-large` to bypass fluid calculation. Ref: https://github.com/WordPress/gutenberg/pull/42757
* Remove the font-size for `post-title` block to inherit based on the set level. `h2` in single post, `h4` in home-page (addresses this [comment](https://github.com/WordPress/twentytwentythree/pull/94#issuecomment-1225369688))
* Update font size of `post-excerpt` to match figma.

### Screenshots

**Home page:**

| Before | After |
| --- | --- |
| image | image |

**Single post title:**

| Before | After |
| --- | --- |
| image | image |

**Paragraphs:**

| Before | After |
| --- | --- |
| image | image |

### DIFF
--- a/theme.json
+++ b/theme.json
@@ -233,25 +233,29 @@
 			"fontSizes": [
 				{
 					"fluid": {
-						"min": "0.875rem"
+						"min": "0.875rem",
+						"max": "1rem"
 					},
 					"size": "1rem",
 					"slug": "small"
 				},
 				{
 					"fluid": {
-						"min": "1rem"
+						"min": "1rem",
+						"max": "1.125rem"
 					},
 					"size": "1.125rem",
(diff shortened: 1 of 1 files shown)


carolinan:
With the removal of the text above the posts, the post titles on the home would be changed to h3.

mikachan (approved):
Thanks for this! These changes look good to me, the paragraph sizes in particular are looking really good 🎉

PR #113, opened 2022-08-24
Update header and footer spacing

mikachan:
~This PR is to help test out https://github.com/WordPress/gutenberg/pull/43576/~

I've updated this PR so that it should be in a good state to bring in, now that the above GB PR has been merged.

The changes include:

- adjusts root vertical padding so that it's consistent (top and bottom now spacing-40)
- removes the top padding from the header and bottom padding from the footer, as they're no longer needed with the root padding change
- changes the spacing variables in the JSON markup (part of https://github.com/WordPress/twentytwentythree/issues/95)

### DIFF
--- a/parts/header.html
+++ b/parts/header.html
@@ -2,6 +2,6 @@
 <div class="wp-block-group"><!-- wp:group {"align":"wide","style":{"spacing":{"padding":{"bottom":"var(--wp--preset--spacing--60)"}}},"layout":{"type":"flex","justifyContent":"space-between"}} -->
 <div class="wp-block-group alignwide" style="padding-bottom:var(--wp--preset--spacing--60)">
 <!-- wp:site-title /-->
-<!-- wp:navigation {"overlayMenu":"always","layout":{"type":"flex","setCascadingProperties":true,"justifyContent":"right"}} /--></div>
+<!-- wp:navigation {"layout":{"type":"flex","setCascadingProperties":true,"justifyContent":"right"}} /--></div>
 <!-- /wp:group --></div>
 <!-- /wp:group -->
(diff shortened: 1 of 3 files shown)


jasmussen:
How is it feeling? You should be able to have different padding values for all 4 sides and so long as there aren't any additional compounding paddings, it should line up. 

The biggest questionmark, I think, is the max-width. Right now it defaults to the content width because it felt the most sensible for the header. But it could default to wide or full as well. 🤔

mikachan:
Sorry, I ran out of time yesterday to write down any notes after testing this!

It seems to work great with TT3 so far on a smaller resolution, or where the content fills the available screen width. I needed to add a top root padding value, as TT3's was set to 0px (for no specific reason), otherwise, the close button had no top padding (correctly, I believe).

https://user-images.githubusercontent.com/1645628/186618138-b9d3e4ab-1a64-4d30-afb1-f6ac32ff35b6.mov

Currently, the max-width doesn't work well for TT3 at larger resolutions, as the header is set to wide align:

https://user-images.githubusercontent.com/1645628/186619595-81110c3d-3e2d-4be7-b01e-4562139bee66.mov

So in this case, it would be better if it defaulted to `--wp--style--global--wide-size`. It's a shame it can't easily use the available space, I guess this is because of the position: fixed on the overlay.

jasmussen:
Good thoughts. I'll update the value in the PR to --wp--style--global--wide-size, as that does use more of the available space and seems to cater to more common designs. 

In the future, I like to think we can improve this code further, so it more naturally stays in sync with the content width of the header part.

jasmussen:
Changed to wide width.

mikachan:
> Changed to wide width.

This works great!

https://user-images.githubusercontent.com/1645628/186630399-186ae9c7-061a-4879-8f7b-6ccf68fec92c.mov

jasmussen:
Cool! Let's test the core PR and see that nothing regresses in other themes, then we can land it! 🥳

carolinan (approved):

PR #120, opened 2022-08-26
404: fix spacing preset, use the new layout setting on the group.

carolinan:
The 404 page used spacing preset 100, which had no value and was not working. This PR updates the preset to 40.

The template also had two group blocks. One group was removed, and the new constrained layout setting was used on the `<main>` group instead.

_The width of the search was reset from 70% to 100% (of the content width), because in the design, the search is wider than the 404 heading. The 404 heading is much larger in the template than in the design, so I'm not sure which is correct._

### DIFF
--- a/patterns/hidden-404.php
+++ b/patterns/hidden-404.php
@@ -14,8 +14,8 @@
 <p class="has-text-align-center"><?php echo esc_html_x( 'This page could not be found.', 'Message to convey that a webpage could not be found', 'twentytwentythree' ); ?></p>
 <!-- /wp:paragraph -->
 
-<!-- wp:spacer {"height":"var(--wp--preset--spacing--100)"} -->
-<div style="height:var(--wp--preset--spacing--100)" aria-hidden="true" class="wp-block-spacer"></div>
+<!-- wp:spacer {"height":"var(--wp--preset--spacing--40)"} -->
+<div style="height:var(--wp--preset--spacing--40)" aria-hidden="true" class="wp-block-spacer"></div>
 <!-- /wp:spacer -->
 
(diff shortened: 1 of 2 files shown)


mikachan (commented):
The spacing and layout changes are looking good to me.

I prefer the width of the search in this PR, but I realise the designs have changed slightly. We've also added a font size preset just to cater for the large 404 heading, which I think works well with this new width. 

@critterverse, we started with the large font size in the designs, and as we began removing the opinionated styles from the templates, we noted it may be better not to have the large 404 heading as a font size preset (as it was very specific). However, we managed to add it without having any impact on any other sizes. 

So.. could you please help us decide: should we stick with the larger font, which is closer to the original design? Or match the new design, with the smaller heading size?

critterverse:
Probably something I'm doing wrong in testing but I'm seeing "This block has encountered an error and cannot be previewed" message on the Group block. Would you mind sharing screenshots of the before and after?

carolinan:
You need to have the development version of Gutenberg.

critterverse:
Ok, got it and understand the question better now. I think the 404 text should be much smaller to match the size of the Post Title, Archive title, etc used on other templates. It was much larger in TT2 but the smaller type feels more like a "blank canvas" IMO.

patterns/hidden-404.php
```diff
@@ -6,16 +6,16 @@
  */
 ?>
 
-<!-- wp:heading {"className":"has-text-align-center","fontSize":"xx-large"} -->
-<h2 class="has-text-align-center has-xx-large-font-size"><?php echo esc_html_x( '404', 'Error code for a webpage that is not found.', 'twentytwentythree' ); ?></h2>
+<!-- wp:heading {"className":"has-text-align-center"} -->
+<h2 class="has-text-align-center"><?php echo esc_html_x( '404', 'Error code for a webpage that is not found.', 'twentytwentythree' ); ?></h2>
```
mikachan: I think removing the font size here works well.

Should we make this heading a h1 similar to the other templates? This would make the default font size the same as other templates too.

mikachan (approved):
LGTM 🚢

PR #121, opened 2022-08-26
WIP: Update templates to use the new layout setting

### LINKED ISSUE #98: Update layout support
I noticed that with Gutenberg trunk active, the wide setting for the query loop stopped working.
The queries and group blocks need to be updated to use `"layout":{"type":"default"}` and `"layout":{"type":"constrained"} `.

This needs to be tested carefully, and issues need to be reported back upstream.

carolinan:
Update templates to use the new layout setting for container blocks:

`"layout":{"type":"constrained"}` : Inner blocks has wide and full width options
`"layout":{"type":"default"}:` inner blocks do not have wide and full width options
These replace `"layout":{"inherit":true} and false`

_Please feel free to pick a template and add it to the same branch._

Closes https://github.com/WordPress/twentytwentythree/issues/98

### DIFF
--- a/parts/comments.html
+++ b/parts/comments.html
@@ -1,4 +1,4 @@
-<!-- wp:group {"layout":{"inherit":true}} -->
+<!-- wp:group {"layout":{"type":"constrained"}} -->
 <div class="wp-block-group"><!-- wp:comments -->
 <div class="wp-block-comments"><!-- wp:comments-title /-->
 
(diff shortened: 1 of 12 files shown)


bgardner:
@carolinan When this replacement occurs, do you notice that the previously existing class of `is-global-padding` disappears?

bgardner:
On a theme of mine, I am replacing this:

```
<!-- wp:post-content {"layout":{"inherit":true}} /-->
```

With this:

```
<!-- wp:post-content {"layout":{"type":"constrained"}} /-->
```

The result is that the original markup of:
```
<div class="has-global-padding is-layout-constrained entry-content wp-block-post-content">
```

Becomes this:

```
<div class="is-layout-constrained entry-content wp-block-post-content">
```

bgardner:
Created an issue in case this is a bug: https://github.com/WordPress/gutenberg/issues/43673

mikachan:
> When this replacement occurs, do you notice that the previously existing class of is-global-padding disappears?

Yes, I can see the same happening on my local site: `is-global-padding` is removed.

bgardner:
@mikachan Would you mind echoing this on the Gutenberg issue I created? I feel like this might be a significant bug and would love to see if we can get it fixed and into 14.0 release.

mikachan:
I've pushed a few commits that I believe cover all the templates. I saw a few merge conflicts with trunk, but these should also be resolved.

I think this is ready for another review!

carolinan:
These look correct to me.

Issue #133, opened 2022-08-28
Use Cover block instead of Post Featured Image to enhance designs


luminuu:
I've worked on a design in Figma with the provided template, before actually looking at the code. I noticed that for the queries and templates, the Post Featured Image block is used to display images. In my design, I thought of giving the images a gradient overlay like you can do with the Cover block. But with the current setup, this design is not possible.

So I want to kick of a discussion of changing the Post Featured Images to Cover blocks, to enable a wider varierty of designs. Any thoughts?

Here's a preview of my design:
![grafik](https://user-images.githubusercontent.com/7585600/187067322-ea51322b-127d-4bdc-b9a4-0158ea89243a.png)

I also saw that @critterverse has been working on a design using duotones but the images don't really display well there. I think with using a gradient, it would be much more pleasing to the eye: https://github.com/WordPress/twentytwentythree/issues/125

critterverse:
Woah, this looks gorgeous @luminuu!

>So I want to kick of a discussion of changing the Post Featured Images to Cover blocks, to enable a wider varierty of designs.

I think one of the most interesting aspects of this project is trying to work within the restrictions of the Post Featured Image block, which will already be populated with content for a lot of folks. Another approach could be to propose adding something like the overlay functionality to the Post Featured Image block — might be a cool proposal for this tracking list: https://github.com/WordPress/twentytwentythree/issues/30

carolinan:
The cover block can be set up to use the existing featured image, the user does not need to upload any new content.
For anyone viewing the page, it looks like the post thumbnail.
I think the difference is, that the block does not have a link, which can be unexpected.
If the post has no featured image, the block does not show.

It still needs to be tested in a template, of course.
````
<!-- wp:cover {"useFeaturedImage":true,"dimRatio":50,"focalPoint":{"x":"0.35","y":"0.36"},"gradient":"luminous-dusk","isDark":false} -->
<div class="wp-block-cover is-light"><span aria-hidden="true" class="wp-block-cover__background has-background-dim has-background-gradient has-luminous-dusk-gradient-background"></span><div class="wp-block-cover__inner-container"></div></div><!-- /wp:cover -->
```

luminuu:
@critterverse I've now created an issue for this and also linked it in #30: https://github.com/WordPress/gutenberg/issues/43676

mikachan:
[...]

PR #136, opened 2022-08-30
Add additional post/page templates

### LINKED ISSUE #101: Consider removing featured image from single post template
When a theme uses the featured image in its single post template, it runs the risk of creating the "duplicate featured image problem."  This happens when a user also adds the image to the post content itself as shown in the following screenshot:

tt3-duplicate-featured

For Twenty Twenty-Three, the image is just sitting above the post itself.  It's not doing anything special, such as serving as a header background.  Plus, there are not many options for customizing the Post Featured Image block via `theme.json` (and no options at all to style this based on context: archive vs. single).  Having no featured image on single post templates also better matches the post editor to what the user sees on the front end.

An alternative approach might be to add a "No Featured Image" template (though, I'd prefer to flip that around to have a selectable "Featured Image" template and no featured image to be the default).

mikachan:
This PR adds additional templates to offer the following options for single posts and pages:

- No featured image or cover block (by default)
- With a featured image
- With a cover block (that automatically uses the featured image)

It also attempts to make the markup more consistent across the different templates:

- Moves the post meta markup into a template part, to make it easy to reuse across multiple templates
- Updates a couple of layout settings to reduce potential merge conflicts with https://github.com/WordPress/twentytwentythree/pull/121
- Moves the common spacing used around comments to the comments template part (spacing-40)

Addresses https://github.com/WordPress/twentytwentythree/issues/101 & part of https://github.com/WordPress/twentytwentythree/issues/133.

## Screenshots

No featured image:
image

With a featured image:
image

With a cover block:
image

With a cover block and featured image:
image


parts/post-meta.html
```diff
@@ -0,0 +1,37 @@
...
+	<hr class="wp-block-separator alignwide has-css-opacity is-style-wide"/>
+	<!-- /wp:separator -->
+
+	<!-- wp:columns {"align":"wide","style":{"spacing":{"margin":{"top":"var:preset|spacing|40"},"blockGap":"var(--wp--preset--spacing--70)"}},"fontSize":"small"} -->
+	<div class="wp-block-columns alignwide has-small-font-size" style="margin-top:var(--wp--preset--spacing--40)">
+		<!-- wp:column {"style":{"spacing":{"blockGap":"0px"}}} -->
+		<div class="wp-block-column">
+			<!-- wp:group {"style":{"spacing":{"blockGap":"1ch"}},"layout":{"type":"flex"}} -->
+			<div class="wp-block-group">
+				<!-- wp:paragraph --><p>Posted</p><!-- /wp:paragraph --><!-- wp:post-date {"format":"F j, Y"} /-->
```
carolinan: (If we are going to make it translation ready, it needs to be in a pattern)

carolinan:
I am not 100% convinced that the cover block in the single template has a purpose unless the post title is the inner content?
(but then it doesn't match what is seen in the post editor)

parts/post-meta.html
mikachan: Yes! I've made it a pattern now.

mikachan:
> I am not 100% convinced that the cover block in the single template has a purpose unless the post title is the inner content?

Yeah, I think this is a really good point, and I like that this would make it even more different from the default template. I've made this change so we can test it out.

I think it's OK that this doesn't match the layout in the post editor, for now.

critterverse:
Good call on moving the title into the Cover block — that works well and I don't think it's an issue with the Cover block not being visible in the post editor. 

One question about the overlay on the Cover block: am I correct in understanding that the overlay color can be set in theme.json but not the opacity? I'm wondering if we should bump the base opacity to at least 50 so that any cool gradient effects could be slightly more visible by default.

richtabor:
> One question about the overlay on the Cover block: am I correct in understanding that the overlay color can be set in theme.json but not the opacity?

Correct. The opacity can be set in the template though. I think 50 would probably work.

mikachan:
Thanks both! I've set the opacity to 50.

patterns/post-meta.php
```diff
@@ -0,0 +1,45 @@
+<?php
+/**
+ * Title: Post Meta
+ * Slug: twentytwentythree/post-meta
+ * Categories: post meta
```
carolinan: This block pattern category does not exist. It could be registered, or another category can be picked and this can be used as keywords instead.

patterns/post-meta.php
```diff
@@ -0,0 +1,45 @@
...
+	<hr class="wp-block-separator alignwide has-css-opacity is-style-wide"/>
+	<!-- /wp:separator -->
+
+	<!-- wp:columns {"align":"wide","style":{"spacing":{"margin":{"top":"var:preset|spacing|40"},"blockGap":"var:preset|spacing|70"}},"fontSize":"small"} -->
+	<div class="wp-block-columns alignwide has-small-font-size" style="margin-top:var(--wp--preset--spacing--40)">
+		<!-- wp:column {"style":{"spacing":{"blockGap":"0px"}}} -->
+		<div class="wp-block-column">
+			<!-- wp:group {"style":{"spacing":{"blockGap":"1ch"}},"layout":{"type":"flex"}} -->
+			<div class="wp-block-group">
+				<!-- wp:paragraph --><p><?php echo esc_html_x( 'Posted', 'Verb to explain the publication status of a post', 'twentytwentythree' ); ?></p><!-- /wp:paragraph --><!-- wp:post-date {"format":"F j, Y"} /-->
```
carolinan: Does the block break if the date format is made translatable? (untested)

patterns/post-meta.php
mikachan: Ah yeah, thanks. I meant to come back to this! I've changed the category to 'query', as this would always be used alongside a post query. If this doesn't work, maybe 'text' would work? And I've added 'post meta' as keywords 👍

patterns/post-meta.php
mikachan: No, it seems to be fine! I've just pushed a commit that allows this to be translatable, and I can see the format still working correctly both in the editor and on the front end. I'm also still able to change it from the site editor.

patterns/post-meta.php
carolinan: I think query is OK.. it is not a query loop but it can be placed in one to create new blog layouts.

carolinan (approved):

PR #145, opened 2022-08-31
Variation: Add variation Primary

ryelle:
primary

That "image" is actually a cover block with the "Sharp circle…" gradient.

primary

(the red & blue really glow together but they don't touch anywhere in the variation itself 😅)

Edit: I'm open to other names — I was aiming for primary like "primary colors", but I don't know if the fact that one of the colors is called primary might be confusing.

### DIFF
--- a/styles/primary.json
+++ b/styles/primary.json
@@ -1,0 +1,192 @@
+{
+	"title": "Primary",
+	"settings": {
+		"color": {
+			"palette": [
+				{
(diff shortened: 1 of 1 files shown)


beafialho:
Thanks for taking the time and effort to create a style variation! Since we had planned on accepting only a few submissions for final release, we will be closing out submissions that were not selected in order to keep this GitHub project focused for contributors.

PR #155, opened 2022-09-09
Remove font smoothing CSS

### LINKED ISSUE #1: style.css and functions.php are only required to handle font-smoothing

scruffian:
This CSS shouldn't be used, see:
https://usabilitypost.com/2012/11/05/stop-fixing-font-smoothing/
https://ui.dev/rwd/articles/font-smoothing

Fixes https://github.com/WordPress/twentytwentythree/issues/1

Individual style variations can adjust font weights for darker background if necessary.


carolinan (approved):

pbking:
I'm keen to remove these resources.  The only difference between this change and t[he first one opened ](https://github.com/WordPress/twentytwentythree/pull/37)is the tweak to the default font-weight.  There's [sound argument against](https://github.com/WordPress/twentytwentythree/pull/37#issuecomment-1217203554) that though.

Are we OK with the default weight un-smoothed?

Though to be honest, changing the default font-weight (or tweaking it in any other way) is probably better served as a separate discussion.  The change at hand is removing the CSS and all the stuff around loading it which I think should be done regardless of how the default font configuration turns out.

I think we should bring this in.

pbking (approved):

PR #159, opened 2022-09-11
Make all font presets fluid, make h1-6 use var()

### LINKED ISSUE #158: Fluid Typography Question
_(First issue / pull request ever — please forgive me if this has already been discussed in this way! I know there have been threads on some of these changes, but couldn't find a bigger-picture discussion of it.)_

Here's my reference PR: **Make all font presets fluid, make h1-6 use var()**

### The question:
Is there a reason that in theme.json (lines 249-283 and 599-633) that variables and built-in fluidity aren't being consistently used? (See before/after GIF for what I'm seeing—looks like only h2 visibly features fluidity.)

### Thoughts:
My opinion is that it would be good--both for code simplicity and general community values--to use variable presets, the scale feature, and Gutenberg's fluidity settings wherever possible, even if rems and clamps allow more design control and precision. 

-the shipped themes are great entry points to coding in WordPress, and committing to the new best practices will be a better educational tool for developers
-fluidity is the most exciting at the largest sizes, and if it's being touted as one of the key features of the theme, it should be default
-using variables instead of units in "styles": will make creating variations and other customizations simpler and more consistent

Those are my two cents, and thanks for reading!

fluid-twentytwentythree

hiyascout:
Fluidity experiment! Example for issue: **Fluid Typography Question**
Fixes #158 

## Changed lines 249-283
- Gave fluid range to 'large' and 'x-large' presets (fluid had been set to false).
- I used extreme sizes to highlight what I'm talking about (not because I think these particular sizes are right).

## Changed lines 599-633 
- Removed all h1–h6 clamp and standard size customizations.
- Instead, I used presets on all h1–h6 sizes

I focused only on playing with fluidity, so refer to the issue I created if doing it this way breaks something else or conflicts with another priority. 

fluid-twentytwentythree

### DIFF
--- a/theme.json
+++ b/theme.json
@@ -602,7 +602,7 @@
 			},
 			"h2": {
 				"typography": {
-					"fontSize": "var(--wp--preset--font-size--x-large)",
+					"fontSize": "clamp(2.625rem, calc(2.625rem + ((1vw - 0.48rem) * 8.4135)), 3.25rem)",
 					"lineHeight": "1.2"
 				}
 			},
--- a/theme.json
+++ b/theme.json
@@ -598,7 +598,7 @@
 			},
 			"h1": {
 				"typography": {
-					"fontSize": "var(--wp--preset--font-size--xx-large)",
+					"fontSize": "3.625rem",
 					"lineHeight": "1.2"
 				}
 			},
--- a/theme.json
+++ b/theme.json
@@ -613,7 +613,7 @@
 			},
 			"h4": {
 				"typography": {
-					"fontSize": "var(--wp--preset--font-size--medium)"
+					"fontSize": "var(--wp--preset--font-size--large)"
 				}
 			},
 			"h5": {
--- a/theme.json
+++ b/theme.json
@@ -258,7 +258,7 @@
 				{
 					"fluid": {
 						"min": "1rem",
-						"max": "4.125rem"
+						"max": "1.125rem"
 					},
 					"size": "1.125rem",
 					"slug": "medium"
--- a/theme.json
+++ b/theme.json
@@ -250,7 +250,7 @@
 				{
 					"fluid": {
 						"min": "0.875rem",
-						"max": "3rem"
+						"max": "1rem"
 					},
 					"size": "1rem",
 					"slug": "small"
--- a/theme.json
+++ b/theme.json
@@ -625,7 +625,7 @@
 			},
 			"h6": {
 				"typography": {
-					"fontSize": "var(--wp--preset--font-size--small)",
+					"fontSize": "var(--wp--preset--font-size--medium)",
 					"textTransform": "uppercase"
 				}
 			},
--- a/theme.json
+++ b/theme.json
@@ -250,7 +250,7 @@
 				{
 					"fluid": {
 						"min": "0.875rem",
-						"max": "1rem"
+						"max": "3rem"
 					},
 					"size": "1rem",
 					"slug": "small"
@@ -258,7 +258,7 @@
 				{
 					"fluid": {
 						"min": "1rem",
-						"max": "1.125rem"
+						"max": "4.125rem"
 					},
 					"size": "1.125rem",
 					"slug": "medium"
@@ -266,13 +266,18 @@
 				{
 					"size": "1.75rem",
 					"slug": "large",
-					"fluid": false
+					"fluid": {
+						"min": "1.75rem",
+						"max": "15rem"
+					}
 				},
 				{
 					"size": "2.25rem",
 					"slug": "x-large",
-					"fluid": false
-				},
+					"fluid": {
+						"min": "2rem",
+						"max": "10rem"
+					}				},
 				{
 					"size": "10rem",
 					"slug": "xx-large",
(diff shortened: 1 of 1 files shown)


theme.json
```diff
@@ -250,29 +250,34 @@
 				{
 					"fluid": {
 						"min": "0.875rem",
-						"max": "1rem"
+						"max": "3rem"
```
mikachan:
```suggestion
						"max": "1rem"
```

(Original intended size.)

theme.json
```diff
@@ -250,29 +250,34 @@
...
 					},
 					"size": "1rem",
 					"slug": "small"
 				},
 				{
 					"fluid": {
 						"min": "1rem",
-						"max": "1.125rem"
+						"max": "4.125rem"
```
mikachan:
```suggestion
						"max": "1.125rem"
```

(Original intended size.)

theme.json
```diff
@@ -593,24 +598,24 @@
 			},
 			"h1": {
 				"typography": {
-					"fontSize": "3.625rem",
+					"fontSize": "var(--wp--preset--font-size--xx-large)",
```
mikachan:
```suggestion
					"fontSize": "3.625rem",
```

I think the XXL size is too large for h1, and I'd prefer to keep the original static size here.

theme.json
```diff
@@ -250,29 +250,34 @@
...
 				},
 				{
 					"size": "2.25rem",
 					"slug": "x-large",
-					"fluid": false
-				},
+					"fluid": {
+						"min": "2rem",
+						"max": "10rem"
+					}				},
```
mikachan:
```suggestion
					"fluid": false
				},
```

I don't think this size preset should be fluid.

theme.json
```diff
@@ -622,7 +627,7 @@
 			},
 			"h6": {
 				"typography": {
-					"fontSize": "var(--wp--preset--font-size--medium)",
+					"fontSize": "var(--wp--preset--font-size--small)",
```
mikachan:
```suggestion
					"fontSize": "var(--wp--preset--font-size--medium)",
```

The small font size is probably too small for a heading, and the h5 and h6 headings were originally designed to be the same font size (the difference in font-weight demonstrates the hierarchy).

theme.json
```diff
@@ -593,24 +598,24 @@
...
 				"typography": {
-					"fontSize": "3.625rem",
+					"fontSize": "var(--wp--preset--font-size--xx-large)",
 					"lineHeight": "1.2"
 				}
 			},
 			"h2": {
 				"typography": {
-					"fontSize": "clamp(2.625rem, calc(2.625rem + ((1vw - 0.48rem) * 8.4135)), 3.25rem)",
+					"fontSize": "var(--wp--preset--font-size--x-large)",
```
mikachan:
```suggestion
					"fontSize": "clamp(2.625rem, calc(2.625rem + ((1vw - 0.48rem) * 8.4135)), 3.25rem)",
```

This heading is already fluid, and using the XL font size makes these headings smaller than the original design.

theme.json
```diff
@@ -593,24 +598,24 @@
...
 					"lineHeight": "1.2"
 				}
 			},
 			"h3": {
 				"typography": {
-					"fontSize": "2.25rem"
+					"fontSize": "var(--wp--preset--font-size--large)"
```
mikachan:
```suggestion
					"fontSize": "var(--wp--preset--font-size--x-large)"
```

I think we should use a variable here, but based on the current sizes on trunk, this would match with XL rather than large.

theme.json
```diff
@@ -593,24 +598,24 @@
...
 				}
 			},
 			"h4": {
 				"typography": {
-					"fontSize": "clamp(1.75rem, calc(1.75rem + ((1vw - 0.48rem) * 8.4135)), 1.875rem)"
+					"fontSize": "var(--wp--preset--font-size--medium)"
```
mikachan:
```suggestion
					"fontSize": "var(--wp--preset--font-size--large)"
```

I think we could use a variable here as well, but again based on the sizes on trunk, this would be large rather than medium.

theme.json
```diff
@@ -250,29 +250,34 @@
...
 					"slug": "medium"
 				},
 				{
 					"size": "1.75rem",
 					"slug": "large",
-					"fluid": false
+					"fluid": {
+						"min": "1.75rem",
+						"max": "15rem"
+					}
```
mikachan:
```suggestion
					"fluid": {
						"min": "1.75rem",
						"max": "1.875rem"
					}
```

This seems like a good preset to make fluid, as we can use this preset for h4 headings too, which were also previously fluid.

mikachan (commented):
Hi @hiyascout! Sorry for the delay on getting back to you with this.

I agree with the feedback you received on https://github.com/WordPress/twentytwentythree/issues/158, in that, not every font size needs to be fluid and it can be overwhelming in certain cases.

However, I really like the idea of using the font size presets more often, where possible. The main reason they're not used for all the heading sizes is that this restricts the design quite a lot. However, I can see from your changes that we could improve where the presets are currently used.

I've made some inline comments, which include changing the font sizes back to their original design. I understand you weren't proposing these sizes and were just using them as examples!

theme.json
hiyascout: @mikachan Wow thank you for these! Question for you or anyone really: Since this is my first foray into this, would I just click 'commit suggestion > commit changes' and then click 'resolve conversation'?

mikachan: Yes, that's right, you can just click 'commit suggestion > commit changes'. That's why I made suggestions instead of just commenting, as it makes it slightly easier for you to make the suggested changes (if you agree with them!)

Alternatively, you can make the changes manually in your code, and then commit and push them as you normally would, if you prefer. These comments will then show as 'outdated', as the code has been updated.

mikachan (approved):
Thanks so much for your contribution and patience here, @hiyascout! I think we should bring this in 👍

Issue #135, opened 2022-08-30
Variation: Summer Night


luminuu:
<img width="498" alt="grafik" src="https://user-images.githubusercontent.com/7585600/187470732-381ff897-a3af-43fa-8332-f1d25158944f.png">

<img width="896" alt="grafik" src="https://user-images.githubusercontent.com/7585600/187470866-2a283a76-8c82-4977-988d-2f1707e17a5b.png">

<img width="990" alt="grafik" src="https://user-images.githubusercontent.com/7585600/187471056-d4388172-128b-4ead-a640-5115da0e3dd8.png">

Figma: https://www.figma.com/file/zv2xtSrcjPQl1AtiVbdX2H/TT3%3A-Summer-Night?node-id=2%3A11

beafialho:
Thanks for taking the time and effort to create a style variation! Since we had planned on accepting only a few submissions for final release, we will be closing out submissions that were not selected in order to keep this GitHub project focused for contributors.

PR #160, opened 2022-09-11
Adds Variation Summer Night

annacmc:
This resolves issue #135 

This is an unfinished draft PR with the basic structure needed to add a new style variation. It needs some more styles added before it will be complete.


mikachan:
Thanks for taking the time and effort to create this JSON file! As this variation wasn't selected to be part of the final release, I'm going to close this PR to help keep this repo focused for contributors.

(I'm hoping we can come back to other variations after the final release, so I've added the Style Variation label so we can easily find this PR later.)

PR #157, opened 2022-09-11
Variation: Grapes

WBerredo:
### Patterns

---

Fixes #126

PR #166, opened 2022-09-14
add pilgrimage theme styles

### LINKED ISSUE #134: Variation: Pilgrimage
grafik

grafik

grafik

Figma: https://www.figma.com/file/0NcNCYScCtXrgbUQ6Z6fAf/TT3%3A-Pilgrimage

kathryncodes:
addresses #134 

Please let me know if any changes are needed and I will be sure to get them in the next couple days. Thanks!

### DIFF
--- a/styles/pilgrimage.json
+++ b/styles/pilgrimage.json
@@ -12,6 +12,11 @@
 					"gradient": "linear-gradient(180deg, var(--wp--preset--color--secondary) 0%,var(--wp--preset--color--primary) 100%)",
 					"name": "Secondary to Primary",
 					"slug": "secondary-primary"
+				},
+				{
+					"gradient": "linear-gradient(180deg, var(--wp--preset--color--base) 0%,var(--wp--preset--color--primary) 350%)",
+					"name": "Base to Primary",
+					"slug": "base-primary"
 				}
 			],
 			"palette": [
@@ -185,6 +190,9 @@
 					}
 				}
 			}
+		},
+		"color": {
+			"gradient": "var(--wp--preset--gradient--base-primary)"
 		},
 		"elements": {
 			"button": {
--- a/styles/pilgrimage.json
+++ b/styles/pilgrimage.json
@@ -1,7 +1,18 @@
 {
-	"title": "Pilgrimage",
 	"settings": {
 		"color": {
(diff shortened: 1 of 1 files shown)


madhusudhand:
@kathryncodes A style variation doesn't need to duplicate all styles from main `theme.json` because they are combined to produce final results. Include the style definitions that are different from the main theme.json. 
As an example, `color palette`, `core-navigation -> elements -> color` etc.. are the changes this variation can have.

Here is a similar variation definition #157 for your reference.

kathryncodes:
@madhusudhand ah okay, I was actually trying to referencing that file but I suppose I didn't need to include everything in it. This file is what I got from using the block theme plugin on my local environment, but I can trim it down. Would your suggestion be to simply remove any code that's the same as theme.json?

carolinan:
> Would your suggestion be to simply remove any code that's the same as theme.json?

Yes

kathryncodes:
Latest commit has all the duplicate theme.json styles removed from pilgrimage.json

@carolinan @madhusudhand thank you for being so helpful!

mikachan:
Thanks for working on this! ❤️ It's looking great. I've just pushed a few commits to fix up some formatting, colours and to add the variation title. I've also added a subtle background gradient as mentioned in [this comment](https://github.com/WordPress/twentytwentythree/issues/134#issuecomment-1239673196).

beafialho:
I'm not seeing the intended duotone in the images: 

Captura de ecrã 2022-09-19, às 10 43 11

mikachan:
> I'm not seeing the intended duotone in the images:

Oh, good spot! It looks like we were missing the duotone filter. I've added this and applied it to all images.

madhusudhand:
Also, should the background be gradient? I refer single color background `base` from figma.

mikachan:
> Also, should the background be gradient? I refer single color background base from figma.

This was requested in [this comment here](https://github.com/WordPress/twentytwentythree/issues/134#issuecomment-1239673196), but it would be great to know if this was the intended effect!

mikachan (approved):
In order to include this variation in Beta 1, I'm going to merge this now. Please feel free to open any additional PRs to make any changes!

PR #171, opened 2022-09-18
Define Inter font family in base theme

mikachan:
This PR moves the Inter font family definition from the Pitch variation to the base theme, so it's available to all variations.


madhusudhand (approved):
Looks good.

Issue #147, opened 2022-09-01
Variation: "Pumpkin Spice"


colorful-tones:
<h1>🌬️ 🌳  <br>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; 🍂 <br> &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; 🍂 <br>
&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; 🍂<br> <br>
&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; 🍂 🍂</h1>

![2023 local_style-guide_](https://user-images.githubusercontent.com/405912/187821068-b7397fd5-00a3-4398-9767-7d6b318e54cd.png)

![2023 local_](https://user-images.githubusercontent.com/405912/187821090-75705f10-b864-407e-bb3a-a039b0f7fc4a.png)

🎃  [spice-latte.zip](https://github.com/WordPress/twentytwentythree/files/9466440/spice-latte.zip)

critterverse:
This style variation has been selected to be bundled with the Twenty Twenty-Three default theme! @ beafialho and I have made some tweaks to the submission as a part of the curation process to ensure that all the selected variations work together nicely as a collection. Here are the requested changes: 

![Pumpkin Spice](https://user-images.githubusercontent.com/77359364/188936971-854b38db-ba52-4dc5-aa29-d3abf150435c.png)

- We love the details in this submission but noticed some similarity and legibility issues with the palette. Instead, we’re suggesting a light bluish/purple pastel background with a white border around the page.
- The orange accent color could be a rich red
- We’ll probably need to come up with a new name — these color suggestions were loosely inspired by @colin-chadwick's [Dusk variation](https://github.com/WordPress/twentytwentythree/issues/50) in case that inspires any ideas!

No explicit next steps are needed — feel free to update theme.json files where relevant or they will be updated in development. Thanks for the contribution!

colorful-tones:
@critterverse - I've taken a pass at updating the colors and renamed from Pumpkin Spice to Whisper. I'm more than happy to iterate further. See: https://github.com/WordPress/twentytwentythree/pull/154

[...]

PR #173, opened 2022-09-19
fix spacing issues with variations

madhusudhand:
This fixes the following.

Pitch: (adjusted top and bottom global padding from 0 to default)

| Before | After |
| --- | --- |
| image | image |

Whisper: (removed additional white global border)

| Before | After |
| --- | --- |
| image | image |

### DIFF
--- a/styles/pitch.json
+++ b/styles/pitch.json
@@ -234,9 +234,7 @@
 		"spacing": {
 			"blockGap": "var(--wp--preset--spacing--40)",
 			"padding": {
-				"top": "0px",
 				"right": "var(--wp--preset--spacing--70)",
-				"bottom": "0px",
 				"left": "var(--wp--preset--spacing--70)"
 			}
 		},
--- a/styles/whisper.json
+++ b/styles/whisper.json
@@ -432,6 +432,11 @@
 				}
 			}
 		},
+		"border": {
+			"color": "var(--wp--preset--color--tertiary)",
+			"style": "solid",
+			"width": "max(1vw, 0.5rem)"
+		},
 		"spacing": {
 			"padding": {
 				"top": "var(--wp--preset--spacing--40)",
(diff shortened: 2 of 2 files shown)


styles/whisper.json
```diff
@@ -432,11 +432,6 @@
 				}
 			}
 		},
-		"border": {
-			"color": "var(--wp--preset--color--tertiary)",
-			"style": "solid",
-			"width": "1vh 1vw"
-		},
```
mikachan: I believe this is part of the design, as mentioned [here](https://github.com/WordPress/twentytwentythree/issues/147#issuecomment-1239673282).

styles/pitch.json
```diff
@@ -234,9 +234,7 @@
 		"spacing": {
 			"blockGap": "var(--wp--preset--spacing--40)",
 			"padding": {
-				"top": "0px",
 				"right": "var(--wp--preset--spacing--70)",
-				"bottom": "0px",
 				"left": "var(--wp--preset--spacing--70)"
```
mikachan, line 238: Looks good!

styles/whisper.json
madhusudhand: Thanks for pointing. Missed that.
I would rather revert it to a fixed width of 0.5rem instead of vh and vw, because they give uneven left-right and top-bottom spacing. Let me know your thoughts.

mikachan: This makes sense to me 👍 Or perhaps we could make this fluid, using something like `max(1vw, 0.5rem)`?

@colorful-tones @beafialho, tagging you for your thoughts here too.

mikachan (approved):

styles/whisper.json
mikachan: In order to include this in Beta 1, I've gone ahead and used `max(1vw, 0.5rem)`. Please feel free to open another PR to make additional changes!

PR #184, opened 2022-09-20
Add Aubergine variation

### LINKED ISSUE #138: Variation "Aubergine"
Screenshot:
Screenshot 2022-08-31 at 09 35 03

Figma:
https://www.figma.com/file/yFcLsr1Hu4ylSnxH61LDOw/TT3-Variation-%22Aubergine%22?node-id=2%3A9

mikachan:
This PR adds the Aubergine style variation, based on the designs and changes described [here](https://github.com/WordPress/twentytwentythree/issues/138).

I really love the background gradient on this one!

### DIFF
--- a/styles/aubergine.json
+++ b/styles/aubergine.json
@@ -206,7 +206,8 @@
 					}
 				},
 				"typography": {
-					"letterSpacing": "0.09rem"
+					"letterSpacing": "0.09rem",
+					"textTransform": "uppercase"
 				}
 			}
 		},
--- a/styles/aubergine.json
+++ b/styles/aubergine.json
@@ -44,6 +44,78 @@
 		}
 	},
 	"styles": {
+		"blocks": {
+			"core/comment-reply-link": {
+				"elements": {
+					"link": {
+						"color": {
+							"text": "var(--wp--preset--color--primary)"
+						},
+						"typography": {
+							"fontStyle": "italic"
+						}
+					}
+				}
+			},
+			"core/post-author-name": {
+				"elements": {
+					"link": {
+						"color": {
+							"text": "var(--wp--preset--color--primary)"
+						},
+						"typography": {
+							"fontStyle": "italic"
+						}
+					}
+				}
+			},
+			"core/post-content": {
+				"elements": {
+					"link": {
+						"color": {
+							"text": "var(--wp--preset--color--primary)"
+						}
+					}
+				}
+			},
+			"core/post-date": {
+				"typography": {
+					"textTransform": "uppercase"
+				}
+			},
+			"core/post-terms": {
+				"elements": {
+					"link": {
+						"color": {
+							"text": "var(--wp--preset--color--primary)"
+						},
+						"typography": {
+							"fontStyle": "italic"
+						}
+					}
+				}
+			},
+			"core/post-title": {
+				"elements": {
+					"link": {
+						"color": {
+							"text": "var(--wp--preset--color--primary)"
+						}
+					}
+				},
+				"color": {
+					"text": "var(--wp--preset--color--primary)"
+				}
+			},
+			"core/separator": {
+				"border": {
+					"width": "2px"
+				},
+				"color": {
+					"text": "var(--wp--preset--color--primary)"
+				}
+			}
+		},
 		"color": {
 			"gradient": "var(--wp--preset--gradient--secondary-base)"
 		},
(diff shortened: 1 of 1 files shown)


styles/aubergine.json
```diff
@@ -0,0 +1,258 @@
...
+	"settings": {
+		"color": {
+			"gradients": [
+				{
+					"gradient": "linear-gradient(180deg, var(--wp--preset--color--secondary) 0%,var(--wp--preset--color--base) 100%)",
+					"name": "Secondary to Base",
+					"slug": "secondary-base"
+				},
+				{
+					"gradient": "linear-gradient(180deg, var(--wp--preset--color--base) 0 10%, var(--wp--preset--color--secondary) 0% 30%, var(--wp--preset--color--base) 100%)",
```
madhusudhand: first gradient should probably be `min(rem, %)` to limit the height of the background on top. With only percentage value, specially in post page with large content, the separation is really somewhere down and isn't looking great.

image

I tried min(24rem, 10%) it seems to work.

```
linear-gradient(180deg, var(--wp--preset--color--base) 0 min(24rem, 10%), var(--wp--preset--color--secondary) 0% 30%, var(--wp--preset--color--base) 100%)
```

styles/aubergine.json
```diff
@@ -0,0 +1,258 @@
...
+					"fluid": false
+				},
+				{
+					"size": "3.25rem",
+					"slug": "x-large",
+					"fluid": false
+				},
+				{
+					"size": "10rem",
+					"slug": "xx-large",
```
madhusudhand, line 85: add name attribute to show proper label in UI

image

styles/aubergine.json
```diff
@@ -0,0 +1,258 @@
...
+						}
+					}
+				},
+				"typography": {
+					"letterSpacing": "0.09rem"
+				}
+			}
+		},
+		"color": {
+			"gradient": "var(--wp--preset--gradient--base-secondary-base)"
```
madhusudhand: add `no-repeat` to avoid gradient repetition for pages with very less content.

```suggestion
			"gradient": "var(--wp--preset--gradient--base-secondary-base) no-repeat"
```

styles/aubergine.json
mikachan: Great idea!!

styles/aubergine.json
mikachan: Hmm, this should be fixed with https://github.com/WordPress/gutenberg/pull/43794. I can see the correct label here, without a name being defined:

image

styles/aubergine.json
mikachan: Looks good!

critterverse:
This looks so good! One small note that the top border on the Row block in the "Blog alternative" template should also be the brighter pink color (to match the separator color).

beafialho:
This looks really good already! A few comments:
- the site title letter spacing looks good, but it should be uppercase

Captura de ecrã 2022-09-21, às 15 06 54

- post dates should have the same letter spacing as the site title

Captura de ecrã 2022-09-21, às 15 08 11

- post dates should have the following format on single posts

Captura de ecrã 2022-09-21, às 15 10 36

- post title on Single post template not looking like the design. Should be the same as the heading on the homepage

Captura de ecrã 2022-09-21, às 15 11 43

beafialho:
Thanks for working on it @mikachan! The last two bullets haven't been addressed yet, could you confirm?

madhusudhand:
> post title on Single post template not looking like the design. Should be the same as the heading on the homepage

This should be fixed here #191

mikachan:
Sorry, I was still working on this! I think I've got this as close as possible using global styles. The one thing I'm struggling with is the colour of the post title block. As this needs to be `primary` in the query loop, I've moved this style to link elements within the query block, however, this then makes the post date links `primary` as well (just in the query block).

image

I've tried overriding this by setting the link element colour within the post date block and by only setting `primary` on `h3` within the query block, but both of these settings are overridden by the link element colour (either of the query block or the default link element).

beafialho:
> Sorry, I was still working on this!

No problem! Sorry if I was rushing you.

> The one thing I'm struggling with is the colour of the post title block.

I think that looks good in orange on the query loop 👍

mikachan:
No worries, I was spending way too long trying to override overrides between different blocks.

Thanks, I'll bring this in!

PR #187, opened 2022-09-20
Styling tweaks to "Block out" variation

critterverse:
Thanks for adding Block out @mikachan! I made a few small tweaks in this PR:

• Removed the hyphen in the variation title (so it will display in the style variation tile as "Block out")
• Reduced the size of the Post Title

I was also wondering if you could help me reduce the line height for the XXL typography used on the Site Title? I wasn't totally sure how it works with the clamp sizes but ideally it would look more like this at the largest size:

| Before  | After |
| ------------- | ------------- |
|  |  |

Thank you!!

### DIFF
--- a/styles/block-out.json
+++ b/styles/block-out.json
@@ -138,7 +138,8 @@
 			},
 			"core/site-title": {
 				"typography": {
-					"fontSize": "var(--wp--preset--font-size--xx-large)"
+					"fontSize": "var(--wp--preset--font-size--xx-large)",
+					"lineHeight": "1.1"
 				}
 			}
 		},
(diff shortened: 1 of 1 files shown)


mikachan:
This is looking good to me!

I've just pushed a commit that adjusts the line height for the site title, although I'm not sure if I picked the correct size at 1.1, but you can see where to edit it now. The default (from the base theme) is set to 1.4, so I imagine you'll need something between 1-1.2.

critterverse:
Line height looks way better now, thanks for fixing @mikachan! This LGTM 🚢

mikachan (approved):

PR #111, opened 2022-08-24
Update home template

mikachan:
This PR removes a section from the home template and tidies up the spacing, based on recent spacing preset changes:

- removes the 'Latest Posts' and 'View more' section, as discussed here: https://github.com/WordPress/twentytwentythree/issues/43#issuecomment-1222063777
- makes the default font-weight of all headings 400, based on the latest designs (and removes this opinion from the home template)
- adjusts the spacing value of the main heading and spacer block
- changes the spacing variable format in the JSON markup, as part of https://github.com/WordPress/twentytwentythree/issues/95)

PR #191, opened 2022-09-22
fix spacing in query-loop and post-title size in single/post templates

### LINKED ISSUE #190: Fix spacing in Home template's query loop
Currently, the home template's query loop spacing is looking way too large. This should be matched with the intended design:

| Local site      | Intended design |
| ----------- | ----------- |
|   Captura de ecrã 2022-09-21, às 16 51 15  |  Captura de ecrã 2022-09-21, às 16 54 29    |

madhusudhand:
This change fixes

* spacing around post title in homepage query loop. 
* adjusts the post-title to `h4` in query loop
* adjusts post title to `h2` in single and post templates.

| Before | After |
| --- | --- |
| image | image |

Fixes: #190

### DIFF
--- a/templates/home.html
+++ b/templates/home.html
@@ -15,7 +15,7 @@
 	<div class="wp-block-query alignwide">
 		<!-- wp:post-template {"align":"wide"} -->
 			<!-- wp:post-featured-image {"isLink":true,"width":"100%","height":"max(15vw, 30vh)","align":"wide"} /-->
-			<!-- wp:post-title {"level":4,"isLink":true} /-->
+			<!-- wp:post-title {"level":3,"isLink":true} /-->
 			<!-- wp:post-excerpt /-->
 			<!-- wp:post-date {"format":"F j, Y","isLink":true} /-->
 
--- a/templates/home.html
+++ b/templates/home.html
@@ -15,7 +15,7 @@
 	<div class="wp-block-query alignwide">
 		<!-- wp:post-template {"align":"wide"} -->
 			<!-- wp:post-featured-image {"isLink":true,"width":"100%","height":"max(15vw, 30vh)","align":"wide"} /-->
-			<!-- wp:post-title {"level":3,"isLink":true} /-->
+			<!-- wp:post-title {"level":4,"isLink":true} /-->
 			<!-- wp:post-excerpt /-->
 			<!-- wp:post-date {"format":"F j, Y","isLink":true} /-->
 
--- a/templates/page.html
+++ b/templates/page.html
@@ -4,7 +4,7 @@
 <main class="wp-block-group">
 	<!-- wp:group {"layout":{"type":"constrained"}} -->
 	<div class="wp-block-group">
-		<!-- wp:post-title {"level":1,"align":"wide","style":{"spacing":{"margin":{"top":"var:preset|spacing|70","bottom":"var:preset|spacing|70"}}}} /-->
+		<!-- wp:post-title {"level":2,"align":"wide","style":{"spacing":{"margin":{"top":"var:preset|spacing|70","bottom":"var:preset|spacing|70"}}}} /-->
 	</div>
 	<!-- /wp:group -->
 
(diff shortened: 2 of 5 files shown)


madhusudhand:
> adjusts the post-title to h4 in query loop

@mikachan By referring designs h4 is the right size, but this changes seems to reverse the change done in #111. 
I might be missing the context, can you confirm if it was intentional?

mikachan:
> By referring designs h4 is the right size, but this changes seems to reverse the change done in https://github.com/WordPress/twentytwentythree/pull/111.

These titles were changed from h4 to h3 to maintain the correct heading hierarchy, as the 'Latest Posts' h3 heading was [removed here](https://github.com/WordPress/twentytwentythree/pull/111/files#diff-6d73f6bbcc50527a0567c49995acacf9cdf1e28ea880751d352a5a914b87e93dL19). 

We shouldn't really be changing the heading level just for styling reasons. Perhaps we need to reduce the font size of these headings via the query block settings instead?

templates/home.html
```diff
@@ -15,7 +15,7 @@ <h2 class="alignwide" style="margin-top:var(--wp--preset--spacing--60);margin-bo
 	<div class="wp-block-query alignwide">
 		<!-- wp:post-template {"align":"wide"} -->
 			<!-- wp:post-featured-image {"isLink":true,"width":"100%","height":"max(15vw, 30vh)","align":"wide"} /-->
-			<!-- wp:post-title {"level":3,"isLink":true} /-->
+			<!-- wp:post-title {"level":4,"isLink":true} /-->
```
mikachan: I think this should stay as a h3, based on [this comment](https://github.com/WordPress/twentytwentythree/pull/191#issuecomment-1254887001).

mikachan (commented):
Thanks for this! The spacing change on the post title block looks good.

This PR has made me reconsider how we're using heading tags across the site. After looking at each template, I don't think we should be using a h1 for the site title in the header (it should just be a paragraph), and instead, ensure that each template has a contextual h1 (e.g. the query title on the archive page is a h1). What do you think?

madhusudhand:
> I don't think we should be using a h1 for the site title in the header (it should just be a paragraph), and instead, ensure that each template has a contextual h1

I Agree. It isn't contextually right to use h1 for Site title, it can just be a paragraph and bold font to be given by its class, where as for H1, the archive page title gives the right context.

templates/page.html
```diff
@@ -4,7 +4,7 @@
 <main class="wp-block-group">
 	<!-- wp:group {"layout":{"type":"constrained"}} -->
 	<div class="wp-block-group">
-		<!-- wp:post-title {"level":1,"align":"wide","style":{"spacing":{"margin":{"top":"var:preset|spacing|70","bottom":"var:preset|spacing|70"}}}} /-->
+		<!-- wp:post-title {"level":2,"align":"wide","style":{"spacing":{"margin":{"top":"var:preset|spacing|70","bottom":"var:preset|spacing|70"}}}} /-->
```
mikachan:
```suggestion
		<!-- wp:post-title {"level":1,"align":"wide","style":{"spacing":{"margin":{"top":"var:preset|spacing|70","bottom":"var:preset|spacing|70"}}}} /-->
```

If we're going to make the site title a paragraph, I think these headings should stay as h1s.

templates/single.html
```diff
@@ -4,7 +4,7 @@
 <main class="wp-block-group">
 	<!-- wp:group {"layout":{"type":"constrained"}} -->
 	<div class="wp-block-group">
-		<!-- wp:post-title {"level":1,"align":"wide","style":{"spacing":{"margin":{"top":"var:preset|spacing|70","bottom":"var:preset|spacing|70"}}}} /-->
+		<!-- wp:post-title {"level":2,"align":"wide","style":{"spacing":{"margin":{"top":"var:preset|spacing|70","bottom":"var:preset|spacing|70"}}}} /-->
```
mikachan:
```suggestion
		<!-- wp:post-title {"level":1,"align":"wide","style":{"spacing":{"margin":{"top":"var:preset|spacing|70","bottom":"var:preset|spacing|70"}}}} /-->
```

mikachan (approved):

carolinan:
1+ for only having one H1 on each page. The home page also needs an H1 though.

PR #196, opened 2022-09-23
Variation Block out: fix font and color issues

### LINKED ISSUE #192: Variation Block-out: font and color issues
Block out variation has the following issues.

* Site title grows very big (to 320px) in large screens due to fluid max 20rem. Limit to 14rem instead.
* Header 2 used in normal context inherits white color. Move the color to specific blocks from `h2` in theme.json

image

* duotone is not applied to home page query block (frontend only, editor works)

image

madhusudhand:
This fixes the following

* limit the max size of the xx-large font to 14rem
* remove color for `h2` and set it to `post-title`
* remove explicit font-size setting to post-title, it inherits the right size with the fix #191 
* add duotone for featured-image

Removing the color for `h2` turns the heading in the homepage to `contrast` color which is expected. However, the design suggests the color to be `primary` which cannot be set in a variation.

Frontend:
image

Design:
image

Fixes: #192

### DIFF
--- a/styles/block-out.json
+++ b/styles/block-out.json
@@ -73,7 +73,7 @@
 				},
 				{
 					"fluid": {
-						"max": "20rem",
+						"max": "14rem",
 						"min": "4rem"
 					},
 					"size": "10rem",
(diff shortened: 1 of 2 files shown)


mikachan:
Thanks for working on this!

> Removing the color for h2 turns the heading in the homepage to contrast color which is expected. However, the design suggests the color to be primary which cannot be set in a variation.

This may be fixed if we change the main heading on the homepage to a h1, as we could set the colour of h1s to primary in the variation.

madhusudhand:
> This may be fixed if we change the main heading on the homepage to a h1, as we could set the colour of h1s to primary in the variation.

fixed here a6a39b9ed691549f446ba9f969672f4617b32599

mikachan (approved):
LGTM!

critterverse:
LGTM too, thank you!

PR #197, opened 2022-09-23
Variation Sherbet: fix fluid typography and font-size of various blocks

### LINKED ISSUE #194: Variation Sherbet: Fix font size in comments
Update the link font size to small.

image

Also, update fluid typography settings by adding explicit `fluid: false` to non fluid sizes and define both `max` and `min` in other size to avoid size growing outsize limits.

madhusudhand:
This fixes following:

* update fluid typography
* fix typography for comments blocks

| Before | After |
| --- | --- |
| image | image |

* update font-size of navigation
* update casing for post-date

Fixes: #194

### DIFF
--- a/styles/sherbet.json
+++ b/styles/sherbet.json
@@ -54,32 +54,38 @@
 		"typography": {
 			"fontSizes": [
 				{
+					"fluid": false,
 					"size": "0.75rem",
 					"slug": "x-small"
 				},
 				{
 					"fluid": {
-						"min": "0.875rem"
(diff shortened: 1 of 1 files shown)


mikachan (commented):
These changes look good to me!

Just one question - why are we disabling fluid typography on so many font size presets?

madhusudhand:
Updated other fonts to appropriate fluid values.

image

mikachan (approved):

PR #199, opened 2022-09-23
Update heading levels in header and home template

mikachan:
This PR changes:

- Site Title to a paragraph (rather than h1)
- Main heading on the home template to h1 (rather than h2)
- Post titles in query loop on home template to h2 (rather than h3), so they follow on correctly from the main heading as h1

Most templates use h1 for their main title (e.g. the query title on the archive template), so with this PR I'm trying to bring the home template in line with other templates.

Also, on reflection, I don't believe having the site title as a h1 is good practice, as this shouldn't be the main title for every page. It also doesn't make sense to have the same h1 title on every page.

Super open to any feedback here, and I realise this change may introduce a need to change the font size of h1s in many/all variations.


carolinan (approved):
I agree that this is the right call.
Every template still needs a quick test, for example, on a single post with no comments, but with comments enabled, the heading order is
H1 -post title
H3-  Leave a reply

With comments:
H1- post title
H2 - 4 responses to post title
H3 leave a reply

carolinan:
I will leave a link to the related pull request for Twenty Twenty-Two, because I do not see a way to change the "Leave a reply" heading in the editor.
https://github.com/WordPress/wordpress-develop/pull/3136

madhusudhand (approved):
This change aligns well with pages such as archive, search where it has h1 header and h2 post-title.

And as mentioned, most of the style variations needs change to adjust the font sizins.

mikachan:
Thanks, both!

Hmm, I might have a fix for the comments headings. I've just pushed a change that adds a h2 above the comments section, which would mean that:

Without comments:
H1- Post title
H2- Comments title
H3- Leave a reply

With comments:
H1- Post title
H2- Comments title
H3- 4 responses to post title
H3- Leave a reply

This does mean that there are two h3's, but they describe sections within the same container section, so I think this works semantically.

With this change, I also moved the comments template into a pattern just to cater for translating the new heading.

carolinan:
In TT2, we ran in to the issue that the comments heading is then showing when comments are off and there are no comments.

mikachan:
> In TT2, we ran in to the issue that the comments heading is then showing when comments are off and there are no comments.

That doesn't seem to be an issue anymore. I've just tested this, and it looks like if comments are turned off, the whole `.wp-block-comments` container is not rendered. This heading sits within this container, so it is also not rendered.

PR #200, opened 2022-09-23
Sherbet: Update colors

scruffian:
This updates the colors on the duotone filter and buttons to match the theme colors. The button effect doesn't work quite so well on this variation as the background of the site doesn't have a fixed background....

Screenshot 2022-09-23 at 16 11 41

Screenshot 2022-09-23 at 16 13 04

### DIFF
--- a/styles/sherbet.json
+++ b/styles/sherbet.json
@@ -163,7 +163,10 @@
 					}
 				},
 				"border": {
-					"radius": "99999px"
+					"color": "var(--wp--preset--color--contrast)",
+					"radius": "99999px",
+					"style": "solid",
+					"width": "2px"
 				},
 				"color": {
 					"background": "var(--wp--preset--color--base)",
--- a/styles/sherbet.json
+++ b/styles/sherbet.json
@@ -7,10 +7,9 @@
 			"duotone": [
 				{
 					"colors": [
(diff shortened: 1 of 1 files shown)


mikachan (approved):
Thanks for this, LGTM!

> The button effect doesn't work quite so well on this variation as the background of the site doesn't have a fixed background....

I know what you mean about the effect, but I think the current setup works well too. I think I prefer the background not to be fixed with this gradient, as I like the gradual rainbow when the page is longer. But I'm happy either way!

@beafialho, would be great to get your thoughts here too 🙏

beafialho:
Is this the intended behaviour of the button? It looks like an error/bug to me. 

Maybe as a way to work this around, the button could have an outline by default or a different gradient to be distinguished from the background.

https://user-images.githubusercontent.com/65220155/192561980-c91c328f-5be5-434e-9489-21f610915905.mp4

mikachan:
I've added an outline to the buttons and tried switching the gradients around between the default button background and the background on hover:


https://user-images.githubusercontent.com/1645628/193253969-85007b78-b986-4258-9310-f047ff282ae3.mov

beafialho:
Thank you @mikachan, I think the button is clearer now.

PR #203, opened 2022-09-24
Bump minimum WP required version

### LINKED ISSUE #24: Increase the minimum required WordPress version
It is common for the default themes to be compatible with WordPress versions below the version the theme is first included in.

This theme's minimum required version is set to 5.9, but the theme can not be activated on 5.9.3:
```
The package could not be installed. The theme is missing the index.php file.
Theme installation failed.
```
The theme could include an index.php file without causing conflicts for 6.1.
But are there other 6.1 features that are breaking on 5.9?
- The spacing CSS variables are not working. So it depends if you consider that breaking or not.
- "Elements" hover and focus styles do not work.
- Font family options do not work.
- The patterns folder does not work on 5.9. So the theme has no footer and no 404 page content.

Updating templates to include new features like the "no results" block for the queries and the new comments blocks also depends on increasing the version.

felipeelia:
Now that WordPress 6.1 Beta 1 was released, it seems we can bump the min required version and close #24

(Sorry for the 1-line PR but I am playing with the theme, and this seemed to be a small but nice change to send. Thanks!)


mikachan (commented):
Brilliant, thank you!

We should also update line 3 of readme.txt. Could you please include that change, and then we'll merge this in?

felipeelia:
Sure thing, @mikachan! Change made in 3b1846c. If any other change is needed, please let me know. Thanks!

mikachan (approved):
Thank you!

PR #204, opened 2022-09-24
Remove "format" attribute from post-date blocks

### LINKED ISSUE #167: Only one post date format is translation ready
In single posts, the post date format is inside a block pattern and uses a translation function:
`<!-- wp:post-date {"format":"<?php echo esc_html_x( 'F j, Y', 'Date format for publication date of post', 'twentytwentythree' ); ?>"} /-->`

In archives/blog/search the post date format can not be translated because it is hard coded:
`	
			
			
			
			
			
			<div style="height:var(--wp--preset--spacing--50)" aria-hidden="true" class="wp-block-spacer"></div>
			
		
`

This means that the date format can end up being different on different parts of the website.
Perhaps it is better to remove the translation function from the pattern, and let the user update it per template, and save the template changes.

felipeelia:
This is a suggestion to close #167

Instead of removing only the translation function, this PR completely removes the format attribute from the post-date blocks so WordPress will fall back to the Date Format set in the Settings Page. Users will still be able to change the format per template.

### DIFF
--- a/patterns/post-meta.php
+++ b/patterns/post-meta.php
@@ -29,7 +29,7 @@
 				</p>
 				<!-- /wp:paragraph -->
 
-				<!-- wp:post-date {"format":"<?php echo esc_html_x( 'F j, Y', 'Date format for publication date of post', 'twentytwentythree' ); ?>"} /-->
+				<!-- wp:post-date /-->
 
 				<!-- wp:paragraph -->
 				<p>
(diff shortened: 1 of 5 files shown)


carolinan (approved):
I like this suggestion, I think it is the best solution.

mikachan (approved):
Thanks for this! I like this solution too 👍

Issue #213, opened 2022-09-28
Style variation Electric: button hover shifts the other contents


tahmidulkarim:
When the button "Get in Touch" is hovered it shifts the contents at the top and bottom of the button.

Here is a video - https://drive.google.com/file/d/1HHDBhe2dP1ApQu3luQBbdfdmu920GSsZ/view

mikachan:
Closed with https://github.com/WordPress/twentytwentythree/pull/215.

PR #215, opened 2022-09-28
Variation Electric: Fix button hover content shift issue in

tahmidulkarim:
This fixes the issue: #213

Added border in the default state so in the hover state the border that is added won't cause the contents to shift anymore.

### DIFF
--- a/styles/electric.json
+++ b/styles/electric.json
@@ -43,6 +43,14 @@
 					"background": "var(--wp--preset--color--contrast)",
 					"text": "var(--wp--preset--color--base)"
 				},
+				"spacing": {
+					"padding": {
+						"top": ".667em",
+						"right": "1.333em",
+						"bottom": ".667em",
+						"left": "1.333em"
+					}
+				},
 				":hover": {
 					"border": {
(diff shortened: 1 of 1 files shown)


styles/electric.json
```diff
@@ -34,6 +34,11 @@
 	"styles": {
 		"elements": {
 			"button": {
+				"border": {
+					"style": "solid",
+					"width": "1px",
```
madhusudhand: 2px is the default for outline button. It is best to define the `2px` border. update the same for hover state.

image

madhusudhand (commented):
Thank you for working on this. 
Posted a comment inline for the border thickness.

styles/electric.json
tahmidulkarim: Thank you for the suggestion @madhusudhand. I have update the border thickness.

madhusudhand (approved):
Pushed a small commit to fix button size issue. The changes are looking good. 👍

mikachan (approved):
Looks good to me too, thank you!

Issue #182, opened 2022-09-20
Ensure all variations include visual change on pseudo states


mikachan:
We should make sure that all variations include a visual change on state change, especially :hover and :active, for accessibility reasons.

- [x] Block-out: button active state does not have good colour contrast https://github.com/WordPress/twentytwentythree/pull/227
- [x] Canary: button hover state is missing https://github.com/WordPress/twentytwentythree/pull/224
- [x] Electric: button active state is missing https://github.com/WordPress/twentytwentythree/pull/228
- [x] Grapes: link hover state is missing https://github.com/WordPress/twentytwentythree/pull/229
- [x] Pilgrimage: button active state is missing https://github.com/WordPress/twentytwentythree/pull/232
- [x] Pitch: button active, hover and focus states do not have good colour contrast https://github.com/WordPress/twentytwentythree/pull/233
- [x] Sherbet: link active state does not have good colour contrast https://github.com/WordPress/twentytwentythree/pull/234
- [x] Whisper: link active state is missing, links in comments have double border on hover https://github.com/WordPress/twentytwentythree/pull/220
- [x] Default style home page and archive: date is missing a hover style. https://github.com/WordPress/twentytwentythree/pull/235
- [x] Default style home page and archive: pagination (older posts / newer posts, but numbers should also be tested) are missing a hover style. https://github.com/WordPress/twentytwentythree/pull/235

folarindeb:
Hi, I would like to help with this issue

mikachan:
Hi @folarindeb, thank you! Please let us know if you have any questions. There's a getting started guide [here in the readme](https://github.com/WordPress/twentytwentythree#getting-started).

RahiDroid:
Hi @mikachan 👋 

If a developer wishes to contribute to this, what would be the right place to look for the design guidelines? I believe current figma files don't include any design guidelines for pseudo-states. Or am I missing something?

mikachan:
Hey @RahiDroid!

Yes, you're right, I don't believe the current figma files include any pseudo-states. I think any contributor should choose style settings they feel work well, and then the changes can be reviewed, with sign-off from @beafialho. The base theme pseudo-state settings, or variations of them, may be worth trying in the first instance and building from there.

carolinan:
- [ ]  Default style home page and archive: date is missing a hover style.
- [ ]  Default style home page and archive: pagination (older posts / newer posts, but numbers should also be tested) are missing a hover style.

[...]

PR #220, opened 2022-09-29
Variation Whisper: fixed links in comments have double border on hover

tahmidulkarim:
Issue #182 

Fixes the last task (Whisper: link active state is missing, links in comments have double border on hover) in the issue.


mikachan (approved):
I've just pushed a change that adds an active state to links, but otherwise, this looks good to me!

PR #222, opened 2022-09-29
Allow no results block content to be translatable

### LINKED ISSUE #217: The "no results" text is not translation ready
In the search.html template, the text inside the query-no-results block "Sorry, but nothing matched your search terms. Please try again with some different keywords" isn't a translatable string.
https://github.com/WordPress/twentytwentythree/blob/trunk/templates/search.html#L31

mikachan:
This PR adds a pattern for the `query-no-results` block content to make it translatable.

Fixes https://github.com/WordPress/twentytwentythree/issues/217.

### DIFF
--- a/patterns/hidden-no-results.php
+++ b/patterns/hidden-no-results.php
@@ -11,4 +11,4 @@
 </p>
 <!-- /wp:paragraph -->
 
-<!-- wp:search {"label":"<?php echo esc_html_x( 'Search', 'label', 'twentytwentythree' ); ?>","placeholder":"<?php echo esc_html_x( 'Search...', 'placeholder for search field', 'twentytwentythree' ); ?>","showLabel":false,"buttonText":"<?php esc_html_e( 'Search', 'twentytwentythree' ); ?>","buttonUseIcon":true} /-->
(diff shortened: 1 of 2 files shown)


patterns/hidden-no-results.php
```diff
@@ -0,0 +1,14 @@
...
+ * Inserter: no
+ */
+?>
+<!-- wp:paragraph -->
+<p>
+<?php echo esc_html_x( 'Sorry, but nothing matched your search terms. Please try again with some different keywords.', 'Message explaining that there are no results returned from a search', 'twentytwentythree' ); ?>
+</p>
+<!-- /wp:paragraph -->
+
+<!-- wp:search {"label":"<?php echo esc_html_x( 'Search', 'label', 'twentytwentythree' ); ?>","placeholder":"<?php echo esc_html_x( 'Search...', 'placeholder for search field', 'twentytwentythree' ); ?>","showLabel":false,"buttonText":"<?php esc_html_e( 'Search', 'twentytwentythree' ); ?>","buttonUseIcon":true} /-->
```
carolinan: One tiny nit, the placeholder and button text are inside HTML attributes (the button text is inside the aria label when the icon is used).

 ```suggestion
<!-- wp:search {"label":"<?php echo esc_html_x( 'Search', 'label', 'twentytwentythree' ); ?>","placeholder":"<?php echo esc_attr_x( 'Search...', 'placeholder for search field', 'twentytwentythree' ); ?>","showLabel":false,"buttonText":"<?php esc_attr_e( 'Search', 'twentytwentythree' ); ?>","buttonUseIcon":true} /-->
```

mikachan: Thank you!

PR #224, opened 2022-09-30
Variation Canary: added hover state on button 

tahmidulkarim:
Issue: #182 

Fixed - Canary: button hover state is missing

Made the button design outline in hover state. Also, added a border around the button in default state so that it won't shift the contents around when the border is added on hover state.


mikachan (approved):
Looking good, thank you!

PR #225, opened 2022-09-30
Remove empty image from home template

### LINKED ISSUE #223: Remove empty image on the home template
The home template includes markup for an image, but no image selected. This prints an empty figure tag also on the front of the website.

https://github.com/WordPress/twentytwentythree/blob/trunk/templates/home.html#L6

mikachan:
This PR swaps the empty image included on the home template for a featured image block.

Fixes https://github.com/WordPress/twentytwentythree/issues/223.


carolinan:
Without testing -does this display the featured image from the first post in the query below it, or? How does it work?

mikachan:
Ah, my intention was that it would use the featured image of a static homepage (if one existed), but I now realise this isn't how the template hierarchy works! We should just remove the image then...

carolinan (approved):

PR #229, opened 2022-09-30
Grapes: Make link hover state different to default state

mikachan:
This updates the link hover state for Grapes, so it's different from the default link state.

Part of https://github.com/WordPress/twentytwentythree/issues/182.

### DIFF
--- a/styles/grapes.json
+++ b/styles/grapes.json
@@ -33,6 +33,17 @@
 	},
 	"styles": {
 		"blocks": {
+			"core/post-comments": {
+				"elements": {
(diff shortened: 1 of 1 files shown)


madhusudhand (commented):
Comments -> Reply button hover state needs a fix.

mikachan:
> Comments -> Reply button hover state needs a fix.

Thanks! Added these here https://github.com/WordPress/twentytwentythree/pull/229/commits/a0441ea0b7b695022d105608d5f32aa0967fba28

PR #235, opened 2022-09-30
Add hover state to Query Pagination and Post Date links

mikachan:
This updates the hover state of the links in Query Pagination and Post Date blocks.

Also tested that the pseudo-state styles work with Query Pagination numbers.

Part of https://github.com/WordPress/twentytwentythree/issues/182.

### DIFF
--- a/theme.json
+++ b/theme.json
@@ -478,31 +478,6 @@
 			"core/comment-edit-link": {
 				"typography": {
 					"fontSize": "var(--wp--preset--font-size--small)"
-				},
-				"elements": {
-					"link": {
-						":hover": {
-							"typography": {
-								"textDecoration": "underline"
-							}
-						},
-						":focus": {
-							"typography": {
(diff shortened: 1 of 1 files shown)


madhusudhand (approved):
The changes looks good.

I have a question. In comments, edit and reply links are in opposite states like 
edit: no-underline (default) and underline (hover)
reply: opposite

Shouldn't they both behave same? I guess "edit" link states are wrong.

carolinan:
I believe Bea made a list of all the different hover states, I'll try to see if I can find it.

mikachan:
Ah, I probably set the styles the wrong way around?  I'll check them against the designs [here](https://www.figma.com/file/ohYEHkZzGjOse46NceTfj5/Twenty-Twenty-Three?node-id=301%3A761).

mikachan:
Thanks both, all the links in the comments should now be consistent.

carolinan (approved):

PR #237, opened 2022-10-03
fix buttion size differences for pitch style variation

### LINKED ISSUE #216: Update the button styles in Pitch style variation
Fill and outline buttons are having different sizes in Pitch style variation.

image

#### Expectation:

Outline button size (padding) should match with fill button.

madhusudhand:
This fixes the outline button size by adding higher specificity for padding.

image

Still the button sizes doesn't match by `2px` because of the border.
To counter, reduced padding by 2px and introduced a border same as background color.

Fixes: #216

### DIFF
--- a/styles/pitch.json
+++ b/styles/pitch.json
@@ -145,7 +145,10 @@
 		"elements": {
 			"button": {
 				"border": {
-					"radius": "0"
+					"radius": "0",
+					"style": "solid",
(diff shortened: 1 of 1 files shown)


beafialho (approved):

mikachan (approved):
Brill, thank you!

PR #239, opened 2022-10-03
Improve block spacing in post meta elements

### LINKED ISSUE #238: The space in the post meta is too large
In the post meta, it looks like the theme is using both spaces inside the text blocks, and block spacing on the parent block:
Screen Shot 2022-10-03 at 12 34 24

mikachan:
This PR slightly reduces the block spacing of child elements in the post meta pattern, from 1ch to 0.5ch.

I tried removing the blockGap size altogether, but this meant it inherited the parent block spacing which is much larger. I then tried setting the block spacing on each nested element to 0px, but this meant all the elements had no spacing at all. Reducing the existing spacing seemed like the best option.

Closes https://github.com/WordPress/twentytwentythree/issues/238.

### DIFF
--- a/patterns/post-meta.php
+++ b/patterns/post-meta.php
@@ -21,7 +21,7 @@
 	<div class="wp-block-columns alignwide has-small-font-size" style="margin-top:var(--wp--preset--spacing--40)">
 		<!-- wp:column {"style":{"spacing":{"blockGap":"0px"}}} -->
 		<div class="wp-block-column">
-			<!-- wp:group {"style":{"spacing":{"blockGap":"1ch"}},"layout":{"type":"flex"}} -->
+			<!-- wp:group {"style":{"spacing":{"blockGap":"0.5ch"}},"layout":{"type":"flex"}} -->
 			<div class="wp-block-group">
 				<!-- wp:paragraph -->
 				<p>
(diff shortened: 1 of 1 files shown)


patterns/post-meta.php
```diff
@@ -57,7 +57,7 @@
 
 		<!-- wp:column {"style":{"spacing":{"blockGap":"0px"}}} -->
 		<div class="wp-block-column">
-			<!-- wp:group {"style":{"spacing":{"blockGap":"1ch"}},"layout":{"type":"flex","orientation":"vertical"}} -->
+			<!-- wp:group {"style":{"spacing":{"blockGap":"0.5ch"}},"layout":{"type":"flex","orientation":"vertical"}} -->
```
colorful-tones, line 60:
```suggestion
			<!-- wp:group {"style":{"spacing":{"blockGap":"0.5ch"}},"layout":{"type":"flex"}} -->
```

colorful-tones (approved):
Tested out the `0.5ch` and looks good. I made one small suggestion for the "Tags:" output to remove `"orientation":"vertical", because this breaks it into two lines.

| before | after |
|---------|------------|
| Screen Shot 2022-10-03 at 5 04 28 PM | Screen Shot 2022-10-03 at 5 04 14 PM |

patterns/post-meta.php
carolinan: The two lines are by design:
https://www.figma.com/file/ohYEHkZzGjOse46NceTfj5/Twenty-Twenty-Three?node-id=301%3A441

colorful-tones: I forgot to reference the design. 🤦  Thanks for clarifying @carolinan

mikachan:
Thanks, @colorful-tones!

PR #244, opened 2022-10-04
Use `esc_attr_x` instead of `esc_html_x` in `hidden-404` pattern.

audrasjb:
This also brings consistency with "Hidden No Results" pattern.


mikachan (approved):
Brill, thank you!

PR #246, opened 2022-10-04
Update copyright years

audrasjb:
This updates readme.txt copyright years to make it consistent with other bundled themes.
See TT2: https://github.com/WordPress/twentytwentytwo/blob/trunk/readme.txt

It also replaces "wordpressdotorg" with "WordPress.org" in the same copyright section, to make it consistent with previous bundled themes as well.


mikachan (approved):
✨ Thank you!

Issue #253, opened 2022-10-05
Blog (Alternative) Template: Only one post displayed on front end


kathrynwp:
In testing the Blog (Alternative) Template, I found that only one post was displayed on the front end, even though all my published posts were displayed in the editor.

### To replicate

1. Publish three posts
2. In Settings > Reading, set "Blog pages show at most" to any number greater than 3
3. Go to Appearance > Editor > Browse all Templates
4. Click the Blog (Alternative) template to open it
5. See three posts are shown in the editor, in the Query Loop block.

https://user-images.githubusercontent.com/7662449/194132118-2ebc0bd9-b4c5-4523-9623-0b4fdaaf7b1c.mp4

6. Create a new page and assign the Blog (Alternative) template to it. Publish the page

### Result

Visit the front end of the new page you just published. Only the most recent post is displayed.

![Testing_Blog_Alternative_–_Twenty_Twenty-Three_Testing](https://user-images.githubusercontent.com/7662449/194132417-85606964-b91a-4f37-863b-5de795337df8.jpg)

### Expected

I expected to see all three published posts on the front end, as I did in the template editor during step 5 above

### Tested with

WordPress 6.1-beta3-54390
Gutenberg 14.2.0
Twenty Twenty-Three downloaded from GitHub approx. 3 hours ago

mikachan:
Fixed by https://github.com/WordPress/twentytwentythree/pull/255.

PR #255, opened 2022-10-06
Blog (Alternative) Template: remove inherit from the query

carolinan:
When `inherit: true` is set on the query loop in the template, the template only shows the page that the template is assigned to.
The PR sets inherit to false, so that the latest 3 posts are displayed.

Closes
https://github.com/WordPress/twentytwentythree/issues/253


mikachan (approved):
Thank you!

PR #258, opened 2022-10-06
Try:  Add border to query pagination

### LINKED ISSUE #205: Remove separator from home template
We should remove this separator from the template. It made sense if there was post navigation but it just feels like it's floating without it.

Captura de ecrã 2022-09-26, às 13 20 28

carolinan:
This is a test where the separator above the query pagination on the home template is removed and replaced with a border. The border and spacing is added to theme.json.

-The border should only be visible if there is a result to paginate through.
You can test this by setting _Reading > Blog pages show at most_ to a number that is higher than the number of posts on  your install.

Downside: The user can't edit or remove the border without editing theme.json.
Upside: Style variations can decide to use a border or not.

Closes https://github.com/WordPress/twentytwentythree/issues/205

### DIFF
--- a/styles/aubergine.json
+++ b/styles/aubergine.json
@@ -180,6 +180,13 @@
 					}
 				}
 			},
+			"core/query-pagination": {
+				"border": {
+					"top": {
+						"color": "var(--wp--preset--color--primary)"
+					}
+				}
+			},
 			"core/separator": {
 				"color": {
 					"text": "var(--wp--preset--color--primary)"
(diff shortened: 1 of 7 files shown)


mikachan (commented):
Nice, I think this should work well!

One issue I can see is that we'll need to match the size of the new border with the separator size for each variation, e.g. Canary:

image

mikachan (approved):
All the variation updates look good to me! ✨

Thanks for working on this!

Issue #218, opened 2022-09-28
Add `accessibility-ready` tag


mikachan:
It would be great if we could add the `accessibility-ready` tag to this theme.

Based on the conversation around adding this tag to TT2 in the [trac ticket here](https://core.trac.wordpress.org/ticket/55172), I believe the tag could also be added to TT3.

Opening this issue in case anything is blocking this (maybe https://github.com/WordPress/twentytwentythree/issues/182, but these should all be fixed soon), and to serve as a reminder to add the tag.

carolinan:
The theme needs to be checked against these requirements:
https://make.wordpress.org/themes/handbook/review/accessibility/required/

mikachan:
I've been through the base theme and each variation, checking the above accessibility requirements using Firefox's accessibility debug tools. I've found the following issues:

### Content Links ([requirements](https://make.wordpress.org/themes/handbook/review/accessibility/required/#content-links))
- [x] Canary (no underline) https://github.com/WordPress/twentytwentythree/pull/259

### Contrasts ([requirements](https://make.wordpress.org/themes/handbook/review/accessibility/required/#contrasts))
- [x] Block out: active link state https://github.com/WordPress/twentytwentythree/pull/264
- [x] Aubergine: active link state https://github.com/WordPress/twentytwentythree/pull/264
- [x] Aubergine: active, focus, hover button states https://github.com/WordPress/twentytwentythree/pull/264
- [x] Marigold: active link state https://github.com/WordPress/twentytwentythree/pull/264
- [x] Marigold: button hover state https://github.com/WordPress/twentytwentythree/pull/264

chakrinya009:
can you please assign this issue

mikachan:
Hi @chakrinya009, thanks for offering to help! All the issues listed here have either been merged already or are included in https://github.com/WordPress/twentytwentythree/pull/264. You could help by testing out https://github.com/WordPress/twentytwentythree/pull/264, if you like.

robglidden:
In RC1, the buttons are nearly unreadable for at least 2 styles - Aubergine and Pilgrimage.

To recreate this I used the Theme Unit Data, and looked at the "Block: Button" post.

The buttons of one of the other darker styles, Pitch, seem reasonable from just looking, though I did not do an accessible contrast check on them.

My guess from inspecting with Chrome devtools and checking the theme.json files is that these styles are using different color combinations (primary, base, contrast, tertiary, secondary etc) for their button styles.

[...]

PR #259, opened 2022-10-06
Canary: Add underlines to links in post content and comment content

carolinan:
Adds underlines to links in post content and comment content blocks. With this PR:

- Default state: solid underline
- Hover: no underline
- Active: no underline
- Focus: dashed underline

Partial for [218.](https://github.com/WordPress/twentytwentythree/issues/218)

### DIFF
--- a/styles/canary.json
+++ b/styles/canary.json
@@ -70,7 +70,7 @@
 	},
 	"styles": {
 		"blocks": {
-			"core/comment-content": {
+			"core/comments": {
 				"elements": {
 					"link": {
 						"typography": {
--- a/styles/canary.json
+++ b/styles/canary.json
@@ -70,6 +70,20 @@
 	},
 	"styles": {
 		"blocks": {
+			"core/comment-content": {
+				"elements": {
+					"link": {
+						"typography": {
+							"textDecoration": "underline"
+						},
+						":hover": {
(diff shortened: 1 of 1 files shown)


styles/canary.json
```diff
@@ -70,6 +70,20 @@
 	},
 	"styles": {
 		"blocks": {
+			"core/comment-content": {
```
mikachan: This doesn't seem to work for me, these styles are being overridden (by the default link styles I think) and I'm seeing no underline by default. 

image

Swapping the block to `core/comments` seems to fix this.

```suggestion
			"core/comments": {
```

carolinan: It is intended for the comment content -Links in text that the comment author adds. That is the area that is covered by the accessibility-ready requirement. But that doesn't seem to be working either.

carolinan: Adding it to core/comments adds it to all links:
Screen Shot 2022-10-06 at 13 56 09

mikachan: Yeah they didn't seem to work for me either, but they were included when I changed the block to `core/comments`.

carolinan: Submitted the change.

mikachan (approved):
Looks good to me!

PR #261, opened 2022-10-06
Block Out variation: Reduce XXL font size

### LINKED ISSUE #251: Block out variation: Site title in footer too large
The site title in the Footer area is extremely large, and looks quite disproportionate in that area:

Post_3_–_Twenty_Twenty-Three_Testing_and_Slack___comet___A8C

Suggest making the Site title in the footer smaller, like it is in the other variations.

mikachan:
This reduces the XXL font size preset in the Block Out variation by around 50%, as discussed [here](https://github.com/WordPress/twentytwentythree/issues/251).

These are before/afters of the maximum size on my screen, at the same browser resolution:

Header:

| Before | After |
| ------ | ----- |
| image | image |

Footer:

| Before | After |
| ------ | ----- |
| image | image |

Closes https://github.com/WordPress/twentytwentythree/issues/251.

cc. @kathrynwp

### DIFF
--- a/styles/block-out.json
+++ b/styles/block-out.json
@@ -73,10 +73,10 @@
 				},
 				{
 					"fluid": {
-						"max": "14rem",
+						"max": "7rem",
 						"min": "4rem"
 					},
-					"size": "10rem",
+					"size": "7rem",
 					"slug": "xx-large"
 				}
 			]
--- a/styles/block-out.json
+++ b/styles/block-out.json
@@ -145,6 +145,12 @@
 				}
 			},
 			"core/site-title": {
+				"spacing": {
+					"padding": {
+						"bottom": "var(--wp--preset--spacing--30)",
+						"top": "var(--wp--preset--spacing--30)"
+					}
+				},
 				"typography": {
 					"fontSize": "var(--wp--preset--font-size--xx-large)",
 					"lineHeight": "1.1",
--- a/styles/block-out.json
+++ b/styles/block-out.json
@@ -159,9 +159,9 @@
 			},
 			"core/query": {
 				"elements": {
-					"h3": {
-						"color": {
-							"text": "var(--wp--preset--color--contrast)"
+					"h2": {
+						"typography": {
+							"fontSize": "var(--wp--preset--font-size--large)"
 						}
 					}
 				}


beafialho:
The size looks good to me.

One thing that could make the **site title** look better is some top and bottom margin to the site title block. I actually tried to set this up in my test site but could not see it on the front end.

| Editor | Front end |
| --- | --- |
| Captura de ecrã 2022-10-06, às 12 41 34 | Captura de ecrã 2022-10-06, às 12 42 04 |

mikachan:
Adding top and bottom didn't seem to work for me either, it looks like it's being overridden by this selector: `body .is-layout-flex > *`. I'll see if there's an upstream issue for this.

I managed to apply the spacing using padding instead. I've pushed up a version of this using the `var(--wp--preset--spacing--30)` preset.

image

carolinan (approved):

mikachan:
Opened an issue here for the margins: https://github.com/WordPress/gutenberg/issues/44736

beafialho:
Thanks!

critterverse:
@mikachan Would you be able to share a screenshot of the home template with these changes? I'm seeing the change to the Site Title but for some reason, am seeing the former size of the Post Titles before changes were made in https://github.com/WordPress/twentytwentythree/pull/187. I'm sure it's just something wrong with my testing environment but just wanted to make sure the type scales work together nicely.

mikachan:
Yes, sure! Here's the home template with these changes:

local-test-site local_

critterverse:
Thanks @mikachan! Weird, what I was seeing in my testing env was correct but I think that we lost the reduction to the Post Title made in https://github.com/WordPress/twentytwentythree/pull/187 somewhere along the way. That PR reduced the Post Title to "large" (so the Post Titles should look significantly smaller than the "Mindblown" typography that appears above and match the size of the navigation). Could we make that change again here?

mikachan:
Ah, we removed the 'large' font size on Post Titles in https://github.com/WordPress/twentytwentythree/pull/196, as it looks like the h2 font size was inheriting the same size, but now it's not. I've set the Post Titles back to 'large' again for this variation:

local-test-site local_

I've done this by setting the font size on the h2s in the query loop, as otherwise the Post Titles in single posts are also set to 'large' as well.

critterverse:
Looks great, thank you!

Issue #4, opened 2022-08-10
Links Details


beafialho:
- [x] Can we keep the `#345C00` color to the links in paragraphs only?

<img width="743" alt="Captura de ecrã 2022-08-10, às 16 51 33" src="https://user-images.githubusercontent.com/65220155/183954204-f251e059-8afd-4597-8913-1308d0d2a03f.png">

- [ ] Also, let's match the links with TT2: `1px` underline thickness and a `text-underline-offset:.25ch` with the same hover effect.

https://user-images.githubusercontent.com/65220155/183954780-6abbeca1-c8c9-4f55-b00b-a5d69d8e2e41.mp4

carolinan:
Was this change implemented? I noticed that the archive block has green links on the front and not in the editor.

mikachan:
> I noticed that the archive block has green links on the front and not in the editor.

I've just noticed that the default link color was using the wrong variable name! Here's a fix: https://github.com/WordPress/twentytwentythree/pull/80

I don't think the underline thickness has been addressed yet. We'd need to do this with CSS.

carolinan:
Should the secondary color be used on links in the paragraph block only? It is on all links in the post content block.

mikachan:
I think the reason the color is set on the post content block is so that links like those in the footer, which are in paragraphs, are not included in the color change.

beafialho:
@mikachan I'd love it if we could improve the appearance of link underlines. The link underlines in the TT3 base theme are currently too thick and close to the word.

- Let's apply `text-underline-offset: .15em` to all links.
- Let's make the underlines thinner as in [this site](https://andersnoren.se/themes/bjork/blog/)

mikachan:
In order to apply the underline offset we'd need to use CSS. I was trying to avoid using any CSS for this theme, to show what's currently possible with Global Styles and theme.json. It also means we don't need to include a functions.php file, as we'd need this to enqueue the stylesheet. Including this file doesn't have any negative impacts from a technical perspective, but I believe not including it allows this theme to be a good example of what can be achieved without this file.

We can change the underline thickness to 1px via theme.json, although without the offset I'm not sure it has the same effect:

<img width="239" alt="image" src="https://user-images.githubusercontent.com/1645628/194302708-2ddd310e-472b-42af-ae6d-86d006e6a5f8.png">

beafialho:
[...]

PR #262, opened 2022-10-06
Base theme: Set all link underlines to 1px

mikachan:
This sets all link underlines on the base theme to 1px thickness.

I think this may be inherited by all the style variations, so if this is approved, we'll need to go through each variation and reset the line thickness in their JSON files.

Part of https://github.com/WordPress/twentytwentythree/issues/4.

### DIFF
--- a/theme.json
+++ b/theme.json
@@ -295,12 +295,12 @@
 					"link": {
 						":hover": {
 							"typography": {
-								"textDecoration": "underline"
+								"textDecoration": "underline 1px"
 							}
 						},
 						":focus": {
 							"typography": {
-								"textDecoration": "underline dashed"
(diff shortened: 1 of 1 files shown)


carolinan:
Chrome, mac, post title block, hover.
Before:
Screen Shot 2022-10-06 at 14 40 55
After:
Screen Shot 2022-10-06 at 14 40 45

carolinan:
Focus:
Screen Shot 2022-10-06 at 14 44 17

carolinan (approved):

beafialho:
Thank you, this change does not look good without the underline offset so I think we should keep it with the previous underline thickness.

PR #269, opened 2022-10-08
Reorder property and fix validation error in json files

t-hamano:
This PR makes the following changes to the global style JSON files:

### Move `$schema` and `version` to top

Related to #54, I think these two properties should be defined first.

### Remove `title` property

Currently, the title property is not allowed in the schema, nor is it used in the site editor.
As [Gutenberg issue](https://github.com/WordPress/gutenberg/issues/44730), it has been suggested that the title be displayed in the preview, but I think that the incorrect property should be removed at this time.

### Unified indentation

In the Marigold theme, tab indentation and space indentation were mixed, so they were unified into tab indentation.

### Remove Incorrect style property

In the Marigold theme, the following incorrect border styles were defined:

```json
"border": {
  "bottom": true
}
```

This property was removed because it doesn't accept a boolean value and doesn't appear to be applied to the style.

### Remaining problems

The code editor doesn't allow the `filter` property, but this is a problem on the JSON schema side and I would like to submit a separate PR to Gutenberg.

filter

If you have any questions about this PR, I would be happy to answer them.

### DIFF
--- a/styles/aubergine.json
+++ b/styles/aubergine.json
@@ -1,5 +1,6 @@
 {
-	"title": "Aubergine",
+	"$schema": "https://schemas.wp.org/trunk/theme.json",
+	"version": 2,
 	"settings": {
 		"color": {
 			"gradients": [
@@ -276,7 +277,5 @@
 		"typography": {
 			"fontFamily": "var(--wp--preset--font-family--dm-sans)"
 		}
-	},
-	"version": 2,
-	"$schema": "https://schemas.wp.org/trunk/theme.json"
+	}
 }
--- a/styles/aubergine.json
+++ b/styles/aubergine.json
@@ -1,6 +1,7 @@
 {
 	"$schema": "https://schemas.wp.org/trunk/theme.json",
 	"version": 2,
+	"title": "Aubergine",
 	"settings": {
 		"color": {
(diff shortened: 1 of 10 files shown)


mikachan (commented):
This is great, thank you for working on it!

> Move $schema and version to top

Makes sense!

> Remove title property

I'd prefer to keep the title properties, as it allows us to format the variation name in Global Styles. Without this property, the file name is used in lowercase:

https://user-images.githubusercontent.com/1645628/194739881-c41ead4e-fe84-4241-976b-883a4df8b81a.mov

> Unified indentation
> Remove Incorrect style property

🙌

> The code editor doesn't allow the filter property

I agree that this should be fixed upstream 👍 It looks like you already opened an issue here: https://github.com/WordPress/gutenberg/issues/44442

t-hamano:
Thank you for the review, @mikachan !

> I'd prefer to keep the title properties, as it allows us to format the variation name in Global Styles. Without this property, the file name is used in lowercase:

Oh, I see that the title property was used. I have restored the title property.
I would like to submit a separate PR to Gutenberg for the issue not defined in the schema 👍

mikachan (approved):
Thank you, looks good to me!

PR #270, opened 2022-10-10
Add accessibility-ready tag

### LINKED ISSUE #218: Add `accessibility-ready` tag

mikachan:
This adds the `accessibility-ready` tag, based on the accessibility check completed as part of https://github.com/WordPress/twentytwentythree/issues/218.

Closes https://github.com/WordPress/twentytwentythree/issues/218.

### DIFF
--- a/style.css
+++ b/style.css
@@ -11,5 +11,5 @@ Version: 1.0
 License: GNU General Public License v2 or later
 License URI: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
 Text Domain: twentytwentythree
-Tags: one-column, custom-colors, custom-menu, custom-logo, editor-style, featured-images, full-site-editing, rtl-language-support, theme-options, threaded-comments, translation-ready, wide-blocks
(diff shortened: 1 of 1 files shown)


robglidden:
Although #218 is closed, I noticed:

In RC1, the buttons are nearly unreadable for at least 2 styles - Aubergine and Pilgrimage.

To recreate this I used the Theme Unit Data, and looked at the "Block: Button" post.

The buttons of one of the other darker styles, Pitch, seem reasonable from just looking, though I did not do an accessible contrast check on them.

My guess from inspecting with Chrome devtools and checking the theme.json files is that these styles are using different color combinations (primary, base, contrast, tertiary, secondary etc) for their button styles.

CleanShot 2022-10-11 at 16 44 09@2x

PR #273, opened 2022-10-11
Pre RC1 merge changes

desrosj:
This makes an adjustment that was included when the theme was synced in RC1 in [Trac-54467](https://core.trac.wordpress.org/changeset/54467).

A few lines in the Marigold style file were indented with spaces instead of tabs.


mikachan (approved):
Thank you!

PR #274, opened 2022-10-11
Add block-patterns tag

kafleg:
Block patterns tag added in style.css file.


mikachan (approved):
Thanks for adding this!

PR #279, opened 2022-10-11
Canary: Apply single font sizes as overrides and remove font family definition on headings

### LINKED ISSUE #278: Canary Variation: Headings element changes not sticking
When using a fresh site using 6.1 Beta 3 and TT3, I noticed that when trying to change the heading element for Canary, the changes wouldn't register: 

https://user-images.githubusercontent.com/26996883/195118706-8a44d4c1-196f-4cb7-b88d-7d21ca5f7ac6.mov

This seems to vary by style variation though as with other variations it works fine!

mikachan:
This changes the way the single font size preset is applied in Canary, as it seems to be causing issues with how typography is used in Global Styles.

Instead of overwriting each font size preset with the same value, this overrides each font size setting to the same preset (`small`).

This also removes the font family setting for `core/heading`, as this doesn't seem to be required anymore and it makes managing the font families from Global Styles less confusing.

Fixes https://github.com/WordPress/twentytwentythree/issues/278.

### DIFF
--- a/styles/canary.json
+++ b/styles/canary.json
@@ -96,11 +96,6 @@
 					"fontSize": "var(--wp--preset--font-size--small)"
 				}
 			},
-			"core/heading": {
-				"typography": {
-					"fontFamily": "var(--wp--preset--font-family--ibm-plex-mono)"
-				}
-			},
 			"core/image": {
 				"border": {
 					"radius": "100px 0 0 0"
--- a/styles/canary.json
+++ b/styles/canary.json
@@ -61,8 +61,7 @@
 				},
 				{
 					"size": "2.25rem",
-					"slug": "x-large",
-					"fluid": false
+					"slug": "x-large"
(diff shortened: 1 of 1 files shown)


styles/canary.json
```diff
@@ -52,19 +52,20 @@
 					"slug": "small"
 				},
 				{
-					"size": "0.75rem",
+					"size": "1.125rem",
```
carolinan, line 55: If the only change to the fontSizes compared to theme.json is that they are not fluid, I wonder if this could be simplified 🤔

mikachan: The difference with the Canary fontSizes is that they're all the same size.

mikachan:
I'm going to bring this in as I believe it addresses the issues from https://github.com/WordPress/twentytwentythree/issues/278, but please open any further related issues if needed.

Issue #280, opened 2022-10-13
Undefined name for XX-Large font size setting


richtabor:
I noticed a missing name value for the `xx-large` font size. 

The others have a value from core (i.e. "Small", "Medium", "Large", Extra Large") — but  `xx-large` is custom to Twenty Twenty Three, so it does not have a `name`. 

![CleanShot 2022-10-13 at 16 01 56@2x](https://user-images.githubusercontent.com/1813435/195698019-894c3c9e-cad0-4880-9047-a121c6527d08.png)

richtabor:
This may be related (odd font size behavior in `Marigold`): 

<img width="478" alt="CleanShot 2022-10-13 at 16 21 27@2x" src="https://user-images.githubusercontent.com/1813435/195702789-7e3dc7be-fb82-4a59-8a45-139e3b3c2290.png">

carolinan:
We are past the string freeze for 6.1, I hope an exception can still be made.

carolinan:
I do find it strange that it does not automatically fall back to for example the slug. I would expect the editor to handle a missing name.

richtabor:
Are the strings in theme.json font size name values translated?

richtabor:
> I would expect the editor to handle a missing name.

Perhaps it should fallback to the `slug` value, if one is not provided.

mikachan:
I think XXL is a Core size, and I believe the 'undefined' issue should have been fixed by https://github.com/WordPress/gutenberg/pull/44791, although I can also currently reproduce this with WP 6.1 RC1.

richtabor:
I don't see it [here](https://github.com/WordPress/gutenberg/blob/trunk/lib/compat/wordpress-6.1/theme.json#L345-L366). Would it be elsewhere?

Yea I'm testing with RC1

mikachan:
The XXL size is listed [here](https://github.com/WordPress/gutenberg/blob/trunk/packages/components/src/font-size-picker/utils.ts#L43).

I've reported the issue [here](https://github.com/WordPress/gutenberg/pull/44791#issuecomment-1280502247), and I'll see if I can find where this has broken between the above issue and now.

mikachan:
I'm now seeing this on all font size presets defined by the theme, not just XXL. It looks like it happens when the theme does not supply a `name` key for the font size preset, even if it's an 'official' Core preset size.

I have a potential fix here: https://github.com/WordPress/gutenberg/pull/45041.

[...]

PR #281, opened 2022-10-13
Add name for xx-large font size

richtabor:
Adds a `name` to the `xx-large` font size, so the UI does not show "undefined."

For #280.

CleanShot 2022-10-13 at 16 03 50@2x


mikachan:
Closing this for now, as this should be fixed as part of the editor changes in WP 6.1 RC2.

PR #290, opened 2022-10-21
Set text colour for file block button

### LINKED ISSUE #288: Consider adding button visited styles
While testing a separate issue, I learnt that the button in the file block has a [default visited text color that is white.](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/file/style.scss#L34)
This makes it inconsistent with other buttons.

Untested, but I think adding a default "visited" style in theme.json would override the block default.

mikachan:
This sets a text colour for the file block button, which overrides a default `:visited` text colour set by [Core here](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/file/style.scss#L34).

Fixes https://github.com/WordPress/twentytwentythree/issues/288.

### DIFF
--- a/styles/aubergine.json
+++ b/styles/aubergine.json
@@ -95,17 +95,6 @@
 	},
 	"styles": {
 		"blocks": {
-			"core/file": {
-				"elements": {
-					"button": {
-						":visited": {
-							"color": {
-								"text": "var(--wp--preset--color--base)"
-							}
-						}
-					}
-				}
-			},
 			"core/comment-reply-link": {
 				"elements": {
 					"link": {
@@ -275,6 +264,11 @@
 						"background": "var(--wp--preset--color--primary)",
 						"gradient": "none",
 						"text": "var(--wp--preset--color--secondary)"
+					}
+				},
+				":visited": {
+					"color": {
+						"text": "var(--wp--preset--color--base)"
 					}
 				}
 			},
--- a/styles/aubergine.json
+++ b/styles/aubergine.json
@@ -95,6 +95,15 @@
 	},
 	"styles": {
 		"blocks": {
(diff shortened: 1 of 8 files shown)


carolinan:
I tested the PR, but only in the default theme.json, not variations. The button text is always black, so when hovering over the button, the text is black over black background.

carolinan:
I think it needs to be
```
			"core/file": {
				"elements": {
					"button": {
						":visited": {
							"color": {
								"text": "var(--wp--preset--color--contrast)"
							}
						}
					}
				}
			},
```
But the schema says it is invalid

mikachan:
Ah that's strange, all the pseudo-states seem to work in the same way as the default button for me. This is what I see on hover:

image

I don't think the `visited` state was included when pseudo-state support was added to theme.json.

carolinan:
It was: https://github.com/WordPress/gutenberg/pull/42096
But was it not added to core then?

carolinan:
I can confirm it is not in the schema and needs to be added.

mikachan:
Ugh I'm sorry, the `:visited` state seems to be working well for me now. I wasn't seeing any new CSS generated for it at all, but I'm guessing it must've been my environment..

I've tested with both the latest GB trunk and just WP 6.1 RC2 (with GB plugin deactivated), and it's working for me in both. Looks like it's just the schema that needs updating for this!

mikachan:
I've moved the `visited` setting to `elements.button` now, for the base theme and all variations. I thought this was cleaner than adding the file block settings.

mikachan:
I'm going to bring this in as I believe it's the fix that @carolinan originally intended, and I'd like this included in RC3.

PR #296, opened 2022-10-27
Try fixing combined hover and visited button states

mikachan:
When both the visited and hover states are active on a button, the visited text colour overrides the hover text colour, which means the two colours are not contrasting in some cases.

This attempts to fix this by adding an !important tag to the hover state text colour. I don't think this is ideal, but re-ordering the properties didn't seem to make a difference here.

Addresses this comment: https://github.com/WordPress/twentytwentythree/issues/288#issuecomment-1293504927


mikachan:
Closing as the theme has been merged into the WordPress Core SVN repository and is no longer maintained on GitHub.

